Issue #169, opened 2014-01-05
Problem installing updated version of Org


joelmccracken:
This is a problem I have run into a few times:

Emacs now comes with a built-in version of Org. However, this is rather old, and I would like to use a more recent version.However,  I can't get cask to install any version of Org using the `depends-on` form. 

I _suspect_ that cask is detecting that org already exists in the load path and is giving up at that point.

IIRC, I have solved this in the past by bypassing Cask, but I would sure like to use cask to manage all of my package dependencies.

alphapapa:
Hitting this issue myself, driving me a bit crazy.  My package requires Org 9, but the built-in version is 8.2.10.  No matter what I try, I can't get Cask to install a current version of Org, so my package's tests fail.

NobbZ:
I do have the same problem, I can't convince cask to install a newer org-mode than what is available from the system-site…

The only way to circumvent this, seems to be to either directly pull from git-repo (I do not want to assume git everywhere) or by specifying `(depends-on "org-plus-contrib" "9.1.0")`.

Can't we have a `(depends-on)` that takes an elpa-repository as `:fetcher`? Or at least http/https to specify release tarballs from github?

NobbZ:
See my transcript:

```plain
~/tmp/foo$ cask --version
0.8.1
~/tmp/foo$ emacs --version
GNU Emacs 24.5.1
Copyright (C) 2015 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
~/tmp/foo$ cask init
~/tmp/foo$ emacs Cask 
~/tmp/foo$ cat Cask
(source org)

(depends-on "org" "9.0.0")
~/tmp/foo$ cask install
~/tmp/foo$ cask eval '(message (org-version))'
8.2.10
~/tmp/foo$ emacs Cask
~/tmp/foo$ cask install
Making version-control local to org-plus-contrib-autoloads.el while let-bound!
Making generated-autoload-file local to  *autoload-file* while let-bound!
Wrote /home/nmelzer/tmp/foo/.cask/24.5/elpa/org-plus-contrib-20180108/org-plus-contrib-autoloads.el
Wrote /home/nmelzer/tmp/foo/.cask/24.5/elpa/org-plus-contrib-20180108/org-loaddefs.el
~/tmp/foo$ cask eval '(message (org-version))'
[...]

PR #232, opened 2014-08-20
Check installed dependency min-version

### LINKED ISSUE #119: org not installable
I am observing a strange behavior. I do not know if this is a bug or a feature. The following `Cask` file does not trigger any install when using `cask install`:

```
(source org)

(depends-on "org")
```

I know that things are working as the following `Cask` files works as expected:

```
(source org)

(depends-on "org-plus-contrib")
```

Would this have anything to do with the fact that a version of `org` is prepackaged with emacs?

marsam:
This fixes #119 and #169

So to install the latest version of org-mode, you can use:

``` elisp
;;; Cask
(package "example.el" "0.0.1" "example package")
(source "org" "http://orgmode.org/elpa/")
(depends-on "org" "8.2.7c")
```

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -480,13 +480,34 @@ returns an `epl-package' object."
         (cask--compute-dependencies dependencies 'cask--find-installed-package)
       dependencies)))
 
+(defun cask--max-version-available (name)
+  "Return the max version available in archives for package with NAME."
+  (let* ((available-pkg (epl-find-available-packages name))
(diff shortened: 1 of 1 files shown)


sambrightman:
If @rejeep will merge then I have a quick test ready to add.

juergenhoetzel:
i would love to see this feature merged. I hit the same issue when trying to install `tramp` from `elpa`.

PR #274, opened 2015-01-13
Fix download command same as 'doc/guide/installation.rst'

syohex:
- Use new domain
- Remove -k(--insecure) option


lunaryorn:
@syohex Thank you!

PR #276, opened 2015-01-16
Add --load-path option

rejeep:
The option takes one argument that is added to Emacs load-path before
the Cask command runs.

### DIFF
--- a/cask-cli.el
+++ b/cask-cli.el
@@ -343,6 +343,10 @@ Commands:
   "Be verbose and do not hide output."
   (setq shut-up-ignore t))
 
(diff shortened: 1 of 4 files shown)


labdsf:
@rejeep Any plan to update this PR and merge it?

I am trying to add `Cask` file to https://github.com/alexott/muse and it stores all `*.el` files in `lisp/` and `contrib/`.

Fuco1:
Closed in favour of #435.  We can revisit this later if the automatic logic proves too intrusive (but I can't imagine a scenario right now where it wouldn't DTRT)

PR #288, opened 2015-04-04
Fix the nil dependencies bug in tar packages.

lxsameer:
By default without providing a `pkg` file for a multi-file package cask will put `nil` in dependencies place at `pkg` file packaged in tar file.

This patch will fix that and put the dependencies in the proper place even without a `pkg` file.

This patch contains several function override of `package-build` tool.

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -904,6 +904,12 @@ NAME-pkg.el or Cask file for the linking to be possible."
   (-when-let (path (cask-dependency-path bundle name))
     (f-symlink? path)))
 
+(defun extract-dependencies (dependencies)
+  "Return a list of (dep-name dep-version)"
+  (mapcar (lambda (x)
+            (list (cask-dependency-name x) (cask-dependency-version x)))
+          dependencies))
+
 (defun cask-package (bundle &optional target-dir)
   "Build an Elpa package of BUNDLE.
 
@@ -912,6 +918,7 @@ a directory specified by `cask-dist-path' in the BUNDLE path."
   (cask--with-package bundle
     (let ((name (symbol-name (cask-bundle-name bundle)))
           (version (cask-bundle-version bundle))
+          (dependencies (extract-dependencies (cask-runtime-dependencies bundle)))
           (patterns (or (cask-bundle-patterns bundle)
                         package-build-default-files-spec))
           (path (cask-bundle-path bundle)))
@@ -919,7 +926,7 @@ a directory specified by `cask-dist-path' in the BUNDLE path."
         (setq target-dir (f-expand cask-dist-path path)))
       (unless (f-dir? target-dir)
         (f-mkdir target-dir))
-      (package-build-package name version patterns path target-dir))))
+      (package-build-package name version patterns path target-dir dependencies))))
 
 (provide 'cask)
 
--- a/cask.el
+++ b/cask.el
@@ -190,6 +190,132 @@ the function `cask--with-environment'.")
 (defconst cask-tmp-packages-path
   (f-expand "packages" cask-tmp-path))
 
+;;;; Package Build Patches
+
+(defun pb/merge-package-info (pkg-info name version dependencies)
+  "Return a version of PKG-INFO updated with NAME, VERSION and info from CONFIG.
+If PKG-INFO is nil, an empty one is created."
+  (let* ((merged (or (copy-sequence pkg-info)
+                     (vector name nil "No description available." version))))
+    (aset merged 0 name)
+    (aset merged 3 version)
+    (when dependencies (aset merged 1 dependencies))
+
+    merged))
+
+(defun package-build-package (package-name version file-specs source-dir target-dir &optional dependencies)
+  "Create PACKAGE-NAME with VERSION.
+
+The information in FILE-SPECS is used to gather files from
+SOURCE-DIR.
+
+The resulting package will be stored as a .el or .tar file in
+TARGET-DIR, depending on whether there are multiple files.
+
+Argument FILE-SPECS is a list of specs for source files, which
+should be relative to SOURCE-DIR.  The specs can be wildcards,
+and optionally specify different target paths.  They extended
+syntax is currently only documented in the MELPA README.  You can
+simply pass `package-build-default-files-spec' in most cases.
+
+DEPENDENCIES is a list of dependencies to override the dependencies 
+extraced from pkg file or single file header.
+
+Returns the archive entry for the package."
+  (when (symbolp package-name)
+    (setq package-name (symbol-name package-name)))
+  (let ((files (package-build-expand-file-specs source-dir file-specs)))
+    (unless (equal file-specs package-build-default-files-spec)
+      (when (equal files (package-build-expand-file-specs
+                          source-dir package-build-default-files-spec nil t))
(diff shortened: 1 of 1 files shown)


lunaryorn:
I'm sorry but I couldn't follow your description.  Could you please try to be more clear?  At best, provide an example that demonstrates the issue?

> This patch contains several function override of package-build tool.

Regardless of the issue at hand this is an absolute no-go, which as far as I'm concerned stands against merging.  If there's something to be fixed, please fix it in the proper place.  And the issue comes from package-build, it should be fixed there.  But I'm strongly against duplicating, let alone overriding package-build functions.

  :-1:

lxsameer:
@lunaryorn For example: 

I have several file which I want to package them.
here is list of them:

```
» ls
Cask  custom.el  inflections.el  kuso-about.el  kuso-base.el  kuso-dpaste.el  kuso-version.el  LICENSE  README.md  session-management.el

```

and here is my cask file:

```elisp
(source gnu)
(source melpa)

(package-file "kuso-base.el")
(package "kuso-base" "1.0.0" "Kuso Base plugin which contain basic configuration of Kuso")

(development
 (depends-on "f")
 (depends-on "ecukes")
 (depends-on "ert-runner")
 (depends-on "el-mock"))

(depends-on "powerline")
(depends-on "color-theme")
(depends-on "key-chord")
(depends-on "color-theme-monokai")
(depends-on "es-lib")
(depends-on "es-windows")
(depends-on "ido-vertical-mode")
;(depends-on "workgroups")
(depends-on "hideshow-org")
(depends-on "dired+")
(depends-on "multiple-cursors")
(depends-on "expand-region")
(depends-on "flx-ido")
```
After I issue a `cask package` a tar package created in the `dist/` directory which contains:

```
» ls
custom.el  inflections.el  kuso-about.el  kuso-base.el  kuso-base-pkg.el  kuso-dpaste.el  kuso-version.el  session-management.el
```

and the content of `kuso-base-pkg.el` is:

```elisp
(define-package "kuso-base" "1.0.0" "Basic configuration of Kuso IDE" 'nil :keywords
  '("lisp" "kuso" "ide" "base"))
;; Local Variables:
;; no-byte-compile: t
;; End:
```

which is not correct. It doesn't contains any dependency information. With this patch content of the `kuso-base-pkg.el` would be:

```elisp
(define-package "kuso-base" "1.0.0" "Basic configuration of Kuso IDE"
  '((flx-ido "")
    (expand-region "")
    (multiple-cursors "")
    (dired+ "")
    (hideshow-org "")
    (ido-vertical-mode "")
    (es-windows "")
    (es-lib "")
    (color-theme-monokai "")
    (key-chord "")
    (color-theme "")
    (powerline ""))
  :keywords
  '("lisp" "kuso" "ide" "base"))
;; Local Variables:
;; no-byte-compile: t
;; End:
```

which is the correct file.

**Regarding to the patch**: I already tried it with `packge-build` guys but they refuse to merge the pull request. they wanted to see how this patch can help `cask`. I'm totally with you on this. This is not a good solution but at least fix the issue and prove my points. 

So when package build guys merged the changes we can get ride of these overrided functions all together.

lunaryorn:
The issue isn't with package-build, it's that Cask doesn't generate a pkg file when packaging.  We should fix that, but meanwhile you can just run `cask pkg-file` before `cask package`.

lunaryorn:
See #289, closing.

lxsameer:
I think generating a `pkg` file is not the write thing to do . **Cask** already knows about the dependencies. why creating an other file with dependencies so package build can find it ? when you can simply and explicitly provide dependencies at the build time.

lxsameer:
Generating a `pkg` file when we know that `package-build` will generate one for us is like move fork around your head before eating. 

Simplicity is better that complexity.

lunaryorn:
@lxsameer No offense meant, but a patch of 135 lines is hardly what I'd call simple, and the code generate a package descriptor from Cask is already there.

We need a package descriptor anyway, so what's the difference in who generates it?  package-build isn't necessarily the best one for this task anyway, since it builds the package descriptor mostly only because it *has* to build a package descriptor to make a valid TAR package.

But that doesn't mean that it's the best or even the right package descriptor.  It's just a best effort in the absence of any better, which is why package-build allows packages to provide their own descriptor.  And as a matter of fact, large packages usually have their own package descriptor already, e.g. [Magit](https://github.com/magit/magit/blob/next/magit-pkg.el).

lxsameer:
@lunaryorn yeah I'm agree with you in the long patch thing, As I mentioned package build guys was to defensive about their code so they wanted to see it in action first ( which I think it's stupid ). But I didn't have other choices.

But about `pkg` file. As you said generating a `pkg` file maybe not the best solution ( which I think it's wrong because of DRY principle ). But why not to implement the better solution ? 

This won't kill anybody or hurt  anyone. It may take more time to implement by at last you'll have the best solution.

lunaryorn:
@lxsameer Respectfully, don't call the package-build maintainers stupid.  Their contributions to the Emacs community outweigh yours (and mine as well), and refusing to merge code whose purpose is unclear is an absolutely understandable position.

Which is why I refused to merge this PR:  It's a *lot* of code, whose advantage over a simple fix is absolutely unclear—other than your claim that it might somehow be “better”.  And we'll eventually have to maintain the code that you dump into our project, even if you are long gone.  That's why I'm wary to merge large code.

I didn't say that generating the package descriptor would not be the best solution.  On the contrary, I *do* think it is.  What I said was that package-build itself is not always in the best position to generate a proper descriptor, which is why it only generates one if none is already present.  Cask is in a better position, so we should make sure that we really use Casks package descriptor for generated packages.

lxsameer:
@lunaryorn don't get me wrong. I didn't call any one stupid. I respect everyone. I just called the idea of putting other package code in this package to prove my point to original authors is stupid. 

And about the explicitly passing dependencies to the `package-build-package` vs creating pkg file before packaging. Let's agree to disagree. With all due respect you guys are too defensive about your code too.

lunaryorn:
@lxsameer Well, as you said, it's *our* code.  Which means that *we* have to maintain it.  In our *own free time*.  I think that gives us the right to be defensive, and to refuse to merge anything which we do not like.  Since in the end it'll be our time spend on maintaining your code…

lxsameer:
@lunaryorn yeah, you're right, your software your rule.

lunaryorn:
@lxsameer Say rather: Our *time*, our rules.

lxsameer:
@lunaryorn what can I say. It's three days that I'm spending my time to fix these kind of bugs but people attitude is not what I was expected. Pointing out bugs and fixing them should not hurt any one. I can't say "our" in such situation

lunaryorn:
@lxsameer I'm sorry that you've spent so long a time on this issue.  Still, we didn't ask you to do so, which implies the risk that your work will be wasted if we disagree.  And you know, you could have asked first, to discuss the issue and possible solutions *before* taking any action and writing 150 lines of code for naught.

lxsameer:
@lunaryorn the code is not important. Even if you folks didn't accept the PR. I had to fork the code and fix it for my needs. Community is all about different ideas and getting defensive about these kind of stuff is not going to help any one.

Issue #289, opened 2015-04-04
Generate pkg-file before packaging


lunaryorn:
Cask needs to run `cask pkg-file` before `cask package`, to make sure that the generated tarball includes all dependencies from the Cask file.  See #288 

/cc @rejeep

lxsameer:
Also your `cask pkg-file` generate the pkg file with wrong dependencies format.

lunaryorn:
@lxsameer In how far is the package descriptor generated by `cask pkg-file` invalid?

lxsameer:
`pkg-file` generate `nil` as the version for dependencies without any version specified which according to `Emacs` docs are wrong. I fixed it in the provided pull request.

lunaryorn:
@lxsameer I see the bug, but I doubt that your fix is the right one.  I'd like to see documentation which states that we should fall back to 0.  As far as I can see, the minimum version is mandatory in `Package-Requires:`.

lxsameer:
I'm sure about it but If you think it's not correct please refer to `define-package` definition in `package.el`. If you don't define the version ( which should be string as said in Emacs define-package docs ) the `version-to-list` can not parse it. 

Later in that function you can see the it simply check the package versions. "0" is the smallest value.

lunaryorn:
@lxsameer Indeed, you are right.  package.el really falls back to `0` for unspecified versions.

lxsameer:
@lunaryorn also `cask` does not include dependencies on single packages too.

lxsameer:
It seems that this issue comes from `package-build` because appereantly they do not check `-pkg.el` file for single package ( because they assume that developer will manually enter the dependencies in single package ) and by convention there is no need for `pkg` files for a single package.

And `package-build` does not provide a way to override this.

lunaryorn:
I'm sorry but I can't follow you.  There's no package descriptor for single
file packages, so what exactly is the issue here?

Am Sonntag, 5. April 2015 schrieb Sameer Rahmani :

lxsameer:
@lunaryorn forget about it. I forgot about `package-file` DSL and how it extract meta data from another file rather than `Cask`.

rejeep:
@lunaryorn Sorry, missed this somehow. It seems that the simplest way to do this would be to simply call `cask-cli/pkg-file` first in `cask-cli/package`. I think testing this would be a bit overkill as we would have to unzip the file and inspect the directory.

PR #290, opened 2015-04-04
pkg file creation function added before package command

lxsameer:
Closing #289

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -730,7 +730,7 @@ If DEEP is t, all dependencies recursively will be returned."
            (-map
             (lambda (dependency)
               (list (cask-dependency-name dependency)
-                    (cask-dependency-version dependency)))
+                    (or (cask-dependency-version dependency) "0")))
             (cask--runtime-dependencies bundle))))
       (pp-to-string `(define-package ,name ,version ,description ',dependencies)))))
 
--- a/cask-cli.el
+++ b/cask-cli.el
@@ -298,6 +298,7 @@ Commands:
 
 (defun cask-cli/package (&optional target-dir)
   "Build package and put in TARGET-DIR or dist if not specified."
+  (cask-cli/pkg-file)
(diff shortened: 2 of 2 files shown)


lunaryorn:
@lxsameer Please use a proper commit message.

cask.el
```diff
@@ -730,7 +730,7 @@ If DEEP is t, all dependencies recursively will be returned."
            (-map
             (lambda (dependency)
               (list (cask-dependency-name dependency)
-                    (cask-dependency-version dependency)))
+                    (or (cask-dependency-version dependency) "0")))
```
lunaryorn: I'm not sure whether I like this.  Generally, you should explicitly specify the minimum version for any runtime dependency.  

I know that we currently don't enforce this in Cask, but that's mostly to avoid inconveniences for those who use Cask for their Emacs configuration.  

@rejeep What do you think?

lxsameer: It's part of Emacs package spec. Dependencies without any package specified should have "0" as version.

lunaryorn: @lxsameer What “Emacs package spec”?

lunaryorn:
@lxsameer I've merged this change.  Thanks for your time and your help.

lxsameer:
no problem. glade to help

cask.el
rejeep: @lunaryorn I think version is more important if it is a package as you say, not so much for Emacs configuration. I think Cask should probably enforce version, as it is the "correct" way of doing it. If your package does not require a specific version, then that package should specify version "0" as a dependency version requirement.

lunaryorn: Idk.  I don't think that there's a “correct” way.  @lxsameer correctly pointed out that this is actually how Emacs itself behaves (it adds "0" for missing versions), so we can just as well do that in Cask, too.

PR #292, opened 2015-05-02
Use the systems temporary file directory

jorissteyn:
I propose this change to make Cask work well when installed with a package manager (i.e. not having write access to the installation directory). Using the real temp directory has other advantages and I couldn't find a reason not to use it by default.


rejeep:
Merged, thanks! :+1:

PR #295, opened 2015-05-19
Support for --emacs command line option.

phillord:
This is just my starter for 10. If it looks okay, then let me know and I will patch up the documentation.

Previously, the emacs executable to be used had to be passed in as an
environment variable, or using heuristics within cask. While this works
under many circumstances, it can be a little erratic when running within
other environments (notably Emacs which can reset the EMACS environment
variable). This option allows the user to be extremely explicit.

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -154,7 +154,9 @@ def get_cask_path(kind):
     separated by path separators.
 
     """
-    process = subprocess.Popen([sys.executable, CASK, kind], stdout=subprocess.PIPE)
+    process = subprocess.Popen([sys.executable, CASK, "--emacs",
+                                get_cask_emacs(),
+                                kind], stdout=subprocess.PIPE)
     stdout, _ = process.communicate()
     return stdout.rstrip()
 
--- a/bin/cask
+++ b/bin/cask
@@ -63,6 +63,9 @@ OSX_EMACSEN = [
     '/usr/local/bin/emacs',
 ]
 
+## An explicitly set Emacs location
+EMACS=None
+
 class UnsupportedEmacsVersionError(Exception):
     """An exception indicating an unsupported Emacs version.
 
@@ -240,10 +243,11 @@ def find_best_emacs():
 def get_cask_emacs():
     """Get the Emacs executable to use for Cask.
 
-    Try to use the command denoted by the environment variable ``$EMACS``.  If
-    the value of this variable does not refer to a real Emacs executable
-    (i.e. Cask is run from inside Emacs), automatically find a good Emacs for
-    the current platform (see ``find_best_emacs``).
+    If the --emacs command line option is given, use that. Otherwise, try to
+    use the command denoted by the environment variable ``$EMACS``. If the
+    value of this variable does not refer to a real Emacs executable (i.e.
+    Cask is run from inside Emacs), automatically find a good Emacs for the
+    current platform (see ``find_best_emacs``).
 
     Return the name or path for the Emacs command Cask shall use as string.
     Raise ``UnsupportedEmacsVersionError``, if the Emacs command does not meet
@@ -254,7 +258,7 @@ def get_cask_emacs():
     # compile.  In this case, $EMACS may be set as well to a rather
     # meaningless value.  If $INSIDE_EMACS is set, we just ignore the
     # value of $EMACS.
-    emacs = ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
+    emacs = EMACS or ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
     emacs = emacs or find_best_emacs()
     ensure_supported_emacs(emacs)
     return emacs
@@ -324,10 +328,21 @@ def main():
             exit_error(
                 'Python 2.6 required, yours is {0.major}.{0.minor}'.format(
                     sys.version_info))
+
+
         if len(sys.argv) > 1 and sys.argv[1] == 'exec':
+            ## never returns!
             exec_command(sys.argv[2:])
-        else:
-            exec_cask(sys.argv[1:])
+
+        ## special handling for --emacs/bin as this has to happen before
+        ## everything else, and we don't want the parsing in two places.
+        if len(sys.argv) > 1:
+            if sys.argv[1] == '--emacs':
+                global EMACS
+                EMACS = sys.argv[2]
+                sys.argv = sys.argv[2:]
+
+        exec_cask(sys.argv[1:])
     except OSError as error:
         exit_error(error)
     except UnsupportedEmacsVersionError as error:
--- a/bin/cask
+++ b/bin/cask
@@ -65,6 +65,7 @@ OSX_EMACSEN = [
 
 ## An explicitly set Emacs location
 EMACS=None
+ARGV=sys.argv
 
 class UnsupportedEmacsVersionError(Exception):
     """An exception indicating an unsupported Emacs version.
@@ -158,6 +159,7 @@ def get_cask_path(kind):
                                 get_cask_emacs(),
                                 kind], stdout=subprocess.PIPE)
     stdout, _ = process.communicate()
+    print( "get_cask_path({}):{}".format(kind, stdout))
     return stdout.rstrip()
 
 
@@ -260,9 +262,22 @@ def get_cask_emacs():
     # compile.  In this case, $EMACS may be set as well to a rather
     # meaningless value.  If $INSIDE_EMACS is set, we just ignore the
     # value of $EMACS.
-    emacs = EMACS or ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
-    emacs = emacs or find_best_emacs()
+    print( "EMACS:{}".format(EMACS))
+    ## command line switch rules everythign
+    if EMACS:
+        emacs = EMACS
+    else:
+        ## if we are inside emacs ignore the EMACS environment because
+        ## it may be set to t.
+        if b'INSIDE_EMACS' not in ENVB:
+            emacs = ENVB.get(b'EMACS')
+
+        ## if we still haven't found anything guess
+        if not emacs:
+            emacs = find_best_emacs()
+
     ensure_supported_emacs(emacs)
+    print( "Returning emacs version: {}".format(emacs))
     return emacs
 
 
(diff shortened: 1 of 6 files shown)


rejeep:
@lunaryorn This is all on you since I don't know any Python.

phillord:
It's not working yet -- some of the consequences of the change are a bit deeper than I had expected. Still comments welcome on why it's not working:-)

bin/cask
```diff
@@ -324,10 +330,21 @@ def main():
...
+            ## never returns!
             exec_command(sys.argv[2:])
-        else:
-            exec_cask(sys.argv[1:])
+
+        ## special handling for --emacs/bin as this has to happen before
+        ## everything else, and we don't want the parsing in two places.
+        if len(sys.argv) > 1:
+            if sys.argv[1] == '--emacs':
+                global EMACS
```
lunaryorn: Don't.  No globals please.  Parse `--emacs` from the command line and pass it on through function arguments, but please don't do “hidden” communication via global variables.

bin/cask
```diff
@@ -324,10 +330,21 @@ def main():
...
         if len(sys.argv) > 1 and sys.argv[1] == 'exec':
+            ## never returns!
             exec_command(sys.argv[2:])
-        else:
-            exec_cask(sys.argv[1:])
+
+        ## special handling for --emacs/bin as this has to happen before
+        ## everything else, and we don't want the parsing in two places.
+        if len(sys.argv) > 1:
+            if sys.argv[1] == '--emacs':
```
lunaryorn: That is pretty flaky.  Please use `argparse` and [`parse_known_args`](https://docs.python.org/2.7/library/argparse.html#partial-parsing) to parse the `--emacs` option.

lunaryorn:
I didn't test it, for it needs a larger refactoring I'll merge it.  Please see my review comments.

bin/cask
phillord: Actually, cask already does that, with `sys.argv` and more subtly in get_cask_path, when the environment is used.

Still, rethreading this is probably better.

phillord:
I look on things the other way around -- it needs testing before it is refactored!

At the moment, the unit tests are trying to load yasnippet-autoload from my (i.e. ~/.emacs.d) configuration file. Very strange, indeed.

bin/cask
lunaryorn: > In bin/cask <https://github.com/cask/cask/pull/295#discussion_r31383712>:
>
> >              exec_command(sys.argv[2:])
> > -        else:
> > -            exec_cask(sys.argv[1:])
> > +
> > +        ## special handling for --emacs/bin as this has to happen before
> > +        ## everything else, and we don't want the parsing in two places.
> > +        if len(sys.argv) > 1:
> > +            if sys.argv[1] == '--emacs':
> > +                global EMACS
>
> Actually, cask already does that, with sys.argv and more subtly in
> get_cask_path, when the environment is used.
>

It's not about the environment, it's about the *global* mutable variable.
I just don't want to maintain code with mutable shared state, and I don't
think that it's good style in a language that has other ways.

phillord:
I am a bit confused -- if I launch 

    ./bin/cask --emacs /usr/local/bin/emacs-24.5 exec ert-runner

I get a crash. The root cause I do not quite understand yet, but superficially, it's because `cask-test/with-sandbox` calls `epl-reset` which runs `package-initialize`, using my own .emacs.d. Does this really make sense?  Is the exec command not running emacs in batch, so it shouldn't be touching my .emacs.d?

bin/cask
phillord: Okay -- I'll need to refactor `exit_error` later on I think, which depends on sys.argv.

Is there a reason for `get_cask_emacs` -- it can be run several times (via `get_cask_path`) and each times requires a full Emacs invocation. Would it not make more sense to set this once an init time?

Would you still be unhappy with the use of a global EMACS variable -- defined once in main, rather similar to the CASK* variables (which are also globals).

lunaryorn: > Is there a reason for get_cask_emacs -- it can be run several times (via get_cask_path) and each times requires a full Emacs invocation. Would it not make more sense to set this once an init time?

I'm not sure what you mean.  `get_cask_path` does not call `get_cask_emacs`.  It runs a new Cask subprocess—which in turn calls `get_cask_emacs` internally, for sure, but would you like to avoid that?

> Would you still be unhappy with the use of a global EMACS variable -- defined once in main, rather similar to the CASK* variables (which are also globals).

The CASK variables are global *constants*.  I'm curious how you'd implement a global `EMACS` variable—presumably storing the path to the selected Emacs instance—as a global constant… unless you move half of `main` into a function to initialize this constant, which I'd not like to see either.

phillord: Sorry, I meant that`get_cask_path` calls `get_cask_emacs` in this PR -- in the trunk of cask this knowledge is based implicitly through the environment. This fails with an --emacs command line.

Yes, I agree that the CASK variables are not altered once set. The EMACS variable would work in the same way. The only question is where the initialization happens. The obvious place being `main()`, yes.

lunaryorn: @phillord Well, it definitely makes sense to call it only once, but the result should not be stored in global state.  As I said, once you have determined the proper Emacs executable, either pass it through via function arguments, or introduce a new class to wrap this state.

> Yes, I agree that the CASK variables are not altered once set. The EMACS variable would work in the same way. 

I'd not…

> The only question is where the initialization happens. The obvious place being main(), yes.

… because the point about the current globals is not that they are set just once.  The point is—and that is why they are *constants* really—that they are set once and only once *at definition*, and at no other place.  And that's the only kind of global a Python program should have in my opinion.

phillord: I think you are making things unnecessarily difficult -- setting at initialization in this case really just means "can be initialized in a single line". I understand why you don't want to pass *changing* global  state around, but the "never changed once initialized" semantics seems a reasonable distinction to me.

Still, your code base, so I will change it that way, once I have fixed the lisp problem.

lunaryorn: > I think you are making things unnecessarily difficult -- setting at initialization in this case really just means "can be initialized in a single line". 

Yes, and that's pretty much how I'd like to keep it.

> I understand why you don't want to pass changing global state around, but the "never changed once initialized" semantics seems a reasonable distinction to me.

I would normally agree, but see, Python does not actually enforce these kind of distinctions, and I think that they are too subtle to rely purely on programmer's discipline for their enforcement.  My point of view is probably distorted by having used languages with strong type systems, but…

> Still, your code base, so I will change it that way, once I have fixed the lisp problem.

… this pretty much nails it.  In the absence of a strong type systems that helps me to catch mistakes, I'd rather prefer to have very strict rules about how variables should be used.

lunaryorn:
@phillord I'd need to see a backtrace and the test output to help you.  Please try passing `--debug` to Cask and ert-runner.

phillord:
With my currently rather hacked version of cask, I am getting the error below from the tests. The root of the problem appears to be when `epl-reset` is run, which runs `package-activate`. There is a `load` form which tries to load the package autoloads which in my case hits yasnippet first -- but it specifically tries to load "yasnippet-autoload" without adding an extension. All inside `with-demoted-errors`, so it shouldn't error, unless debug-on-error has been switched on my ert/ert-runner.

Ah, well, I am a little further forward. Still don't know why this is happening, and will look more later.


    (let ((old-lp load-path)) (condition-case err (load (expand-file-nam
    (let* ((name (progn (or (and (vectorp pkg-desc) (>= (length pkg-desc
    package-activate-1([cl-struct-package-desc yasnippet (0 9 0 1) "Yet 
    (if fail (warn "Unable to activate package `%s'.\nRequired package `
    (let* ((pkg-vec (car pkg-descs)) (fail (catch (quote dep-failure) (l
    (cond ((null pkg-descs) (package-built-in-p package)) ((and (memq pa
    (let ((pkg-descs (cdr (assq package package-alist)))) (while (if pkg
    package-activate(yasnippet)
    (let ((elt (car --dolist-tail--))) (package-activate (car elt)) (set
    (while --dolist-tail-- (let ((elt (car --dolist-tail--))) (package-a
    (let ((--dolist-tail-- package-alist)) (while --dolist-tail-- (let (
    (if no-activate nil (let ((--dolist-tail-- package-alist)) (while --
    package-initialize(nil)
    epl-initialize()
    epl-change-package-dir("~/.emacs.d/elpa")
    epl-reset()
    (unwind-protect (let* ((default-directory cask-test/sandbox-path) (c
    (let ((f--guard-paths paths)) (unwind-protect (let* ((default-direct
    (unwind-protect (let ((f--guard-paths paths)) (unwind-protect (let* 
    (let ((paths (if (listp (list cask-test/sandbox-path cask-test/cvs-r
    (closure (t) nil (let ((paths (if (listp (list cask-test/sandbox-pat
    #[0 "\306\307!r\211q\210\310\311\312\313\314\315!\316\"\317\320%DC
    funcall(#[0 "\306\307!r\211q\210\310\311\312\313\314\315!\316\"\31
    ert--run-test-internal([cl-struct-ert--test-execution-info [cl-struc
    #[0 "r\304 q\210\305 )\306\307\310\311\312\313!\314\"\315\316%DC\2
    funcall(#[0 "r\304 q\210\305 )\306\307\310\311\312\313!\314\"\315\
    ert-run-test([cl-struct-ert-test cask-add-dependency-test/developmen
    ert-run-or-rerun-test([cl-struct-ert--stats (and t) [[cl-struct-ert-
    ert-run-tests((and t) (lambda (event-type &rest event-args) (cond ((
    ert-runner/run-tests-batch((and t))
    (let ((stats (ert-runner/run-tests-batch selector))) (kill-emacs (if
    ert-runner/run-tests-batch-and-exit((and t))
    (if ert-runner-verbose (ert-runner/run-tests-batch-and-exit ert-runn
    (let ((test-files (ert-runner--test-files tests)) (test-helper (f-ex
    ert-runner/run()
    apply(ert-runner/run nil)
    commander--handle-command(nil)
    commander-parse(nil)
    (if commander-parsing-done nil (commander-parse (or commander-args (
    (progn (setq commander-default-config nil) (setq commander-options n
    eval-buffer(#<buffer  *load*> nil "/home/phillord/src/large/cask/.ca
    load-with-code-conversion("/home/phillord/src/large/cask/.cask/24.5.
    load("/home/phillord/src/large/cask/.cask/24.5.1/elpa/ert-runner-201
    command-line-1(("-scriptload" "/home/phillord/src/large/cask/.cask/2
    command-line()
    normal-top-level()

PR #296, opened 2015-05-19
Adds support for eval command.

phillord:
I added this command so I could check which Emacs version was running, but it seems that it might be more generically useful. Will patch up the documentation iff you think its worth while.

This allows executing arbitrary lisp in the context of the current
project.

### DIFF
--- a/cask-cli.el
+++ b/cask-cli.el
@@ -224,6 +224,20 @@ directory called bin in the root directory.
 The output is formatted as a colon path."
   (princ (concat (s-join path-separator (cask-exec-path (cask-cli--bundle))) "\n")))
 
+(defmacro cask-cli--with-package-path (&rest body)
+  "Execute BODY with `load-path' set according to the project."
+  (declare (debug t))
+  `(let ((load-path
+          (cask-load-path (cask-cli--bundle))))
+     (add-to-list 'load-path
+                  cask-cli--path)
+     ,@body))
+
+(defun cask-cli/eval (form)
+  "Eval FORM with the `load-path' set according to the project."
+  (cask-cli--with-package-path
+   (eval (read form))))
+
 (defun cask-cli/package-directory ()
   "Print current package installation directory."
   (princ (concat (cask-elpa-path (cask-cli--bundle)) "\n")))
@@ -365,6 +379,7 @@ Commands:
  (command "help [command]" cask-cli/help)
  (command "load-path" cask-cli/load-path)
  (command "exec-path" cask-cli/exec-path)
+ (command "eval <form>" cask-cli/eval)
  (command "path" cask-cli/exec-path)
  (command "package-directory" cask-cli/package-directory)
  (command "outdated" cask-cli/outdated)
--- a/Cask
+++ b/Cask
@@ -15,4 +15,5 @@
  (depends-on "ert-runner")
  (depends-on "el-mock")
  (depends-on "noflet")
- (depends-on "ert-async"))
+ (depends-on "ert-async")
+ (depends-on "shell-split-string"))
--- a/doc/guide/usage.rst
+++ b/doc/guide/usage.rst
@@ -75,6 +75,24 @@ Execute the system :var:`command` with the given :var:`arguments`, with a
 proper `$PATH` (see :ref:`cask path`) and `$EMACSLOADPATH` (see :ref:`cask
 load-path`).
 
+
+.. _cask eval:
+
+cask eval
+---------
+
+.. program:: cask eval
+
+::
+
+   cask [GLOBAL-OPTIONS] eval [FORM]
+
+Evaluate ``FORM`` as a lisp form with a proper `$PATH` (see :ref:`cask path`)
+and $EMACSLOADPATH (see :ref:`cask load-path`). The return value of the form
+is not printed directly: ``FORM`` must print to the standard out or error
+stream.            
+
+
 .. _cask help:
 
 cask help
--- a/features/eval.feature
+++ b/features/eval.feature
@@ -1,0 +1,18 @@
+Feature: Eval
+  Link local packages
+
+  Scenario: No arguments
+    Given this Cask file:
+      """
+      (package "super-project" "0.0.1" "Super project.")
+      """
+    When I run cask "eval '(print nil)'"
+    Then I should not see command error:
+      """
+      Wrong type argument: stringp, nil
+      """
+    But I should see command output:
+      """
+
+      nil
+      """
--- a/features/step-definitions/cask-steps.el
+++ b/features/step-definitions/cask-steps.el
@@ -34,6 +34,8 @@
   (defvar cask-test/stdout)
   (defvar cask-test/stderr))
 
+(require 'shell-split-string)
+
 (defun cask-test/template (command)
   "Return COMMAND with placeholders replaced with values."
   (->> command
@@ -48,7 +50,7 @@
   (lambda (filename content)
     (f-write-text content 'utf-8 (f-expand filename cask-test/sandbox-path))))
 
-(When "^I run cask \"\\([^\"]*\\)\"$"
+(When "^I run cask \"\\(.*\\)\"$"
   (lambda (command)
     ;; Note: Since the Ecukes tests runs with Casks dependencies in
     ;; EMACSLOADPATH, these will also be available in the subprocess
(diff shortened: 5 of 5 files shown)


cask-cli.el
```diff
@@ -365,6 +370,7 @@ Commands:
  (command "help [command]" cask-cli/help)
  (command "load-path" cask-cli/load-path)
  (command "exec-path" cask-cli/exec-path)
+ (command "eval [*]" cask-cli/eval)
```
rejeep: `[*]` means zero or more. `[foo]` means one or more.

rejeep:
This would be very useful if all project dependencies where in path. So the same as:

```bash
$ cask exec emacs --eval '...' --batch
```

phillord:
Ah, I assumed that cask-cli.el was already running in that environment -- obviously I was wrong!

cask-cli.el
```diff
@@ -224,6 +224,14 @@ directory called bin in the root directory.
 The output is formatted as a colon path."
   (princ (concat (s-join path-separator (cask-exec-path (cask-cli--bundle))) "\n")))
 
+(defun cask-cli/eval (&optional form)
+  (mapc (lambda (n) (add-to-list 'load-path n))
+        (cask-load-path (cask-cli--bundle)))
+  (add-to-list 'load-path ".")
```
rejeep: Use `cask-cli--path` instead of `"."`

cask-cli.el
```diff
@@ -224,6 +224,14 @@ directory called bin in the root directory.
 The output is formatted as a colon path."
   (princ (concat (s-join path-separator (cask-exec-path (cask-cli--bundle))) "\n")))
 
+(defun cask-cli/eval (&optional form)
+  (mapc (lambda (n) (add-to-list 'load-path n))
```
rejeep: How about using dynamic binding?

```lisp
(let ((load-path (cask-load-path (cask-cli--bundle))))
  ;; ...
  )
```

cask-cli.el
```diff
@@ -224,6 +224,14 @@ directory called bin in the root directory.
...
+        (cask-load-path (cask-cli--bundle)))
+  (add-to-list 'load-path ".")
+  (with-temp-buffer
+    (insert form)
+    (eval-buffer)))
```
rejeep: You could use `(eval (read form))`

cask-cli.el
phillord: Thought about it. Didn't bother because it's "cli" --- i.e. emacs is going to terminate anyway, which is as good a way to reset the load-path as a dynamic binding.

I've added it as a macro. If you add "eval", then someone is going to want "eval-file" before long:-)

cask-cli.el
phillord: I knew there was a better way of doing that!

cask-cli.el
```diff
@@ -224,6 +224,20 @@ directory called bin in the root directory.
 The output is formatted as a colon path."
   (princ (concat (s-join path-separator (cask-exec-path (cask-cli--bundle))) "\n")))
 
+(defmacro cask-cli/with-package-path (&rest body)
```
rejeep: We only use the slash functions for commands/options. Can you rename this to: `cask-cli--with-package-path `

rejeep:
Could you please also write an Ecukes test for this? Should be fairly easy!

phillord:
Well, it would be fairly easy to add an ecukes test but I can't work out how to get the step definitions working because the command in question includes escaped strings. This is my best attempt so far.

I am getting ecukes failures anyway, so it's hard to be sure.

cask-cli.el
```diff
@@ -365,6 +379,7 @@ Commands:
  (command "help [command]" cask-cli/help)
  (command "load-path" cask-cli/load-path)
  (command "exec-path" cask-cli/exec-path)
+ (command "eval [form]" cask-cli/eval)
```
rejeep: Thinking about it, the `form` argument is not optional, so it should be: `"eval form"`

rejeep:
Doesn't something like this work?

```cucumber
When I run cask "eval '(print emacs-version)'"
```

phillord:
No. Took me a while to work it out why. I tried

    When I run cask "eval '(print nil)'"

Then I get (er..) a crash. What is happening is that somewhere in the system `'(print` is interpreted as one argument and `nil')` is interpreted as a the second, which results in a "Wrong number of arguments" error.

rejeep:
The errors occurs because it's split on space: https://github.com/cask/cask/blob/master/features/step-definitions/cask-steps.el#L59

phillord:
Okay, so this commit should be slightly closer. But ecukes is still crashing and I do not understand why.

rejeep:
Didn't know about `shell-split-string`, cool.

Because you are initialising the Cask project, you need a Cask file. Adding something like this will work:

```diff
diff --git a/features/eval.feature b/features/eval.feature
index d1f5e31..8227e5e 100644
--- a/features/eval.feature
+++ b/features/eval.feature
@@ -2,13 +2,16 @@ Feature: Eval
   Link local packages
 
   Scenario: No arguments
-    When I run cask "eval '(print nil)'"
+    Given this Cask file:
+      """
+      (package "super-project" "0.0.1" "Super project.")
+      """
+    When I run cask "eval '(print "foo")'"
     Then I should not see command error:
       """
       Wrong type argument: stringp, nil
       """
     But I should see command output:
       """
-
-      nil
+      foo
       """
```

phillord:
`shell-split-string` was a lucky fine indeed. There is just no easy way to do this without.

Tests are running now. I am using `nil` rather than `"foo"` because it does not require quoting and has a predictable outcome (unlike printing a variable). That is worth the risk of a less specific output in my book.

rejeep:
You still have to change `[form]` to `<form>`. It's not optional, it's required.

rejeep:
Can you also squish all commits into a single one.

phillord:
I've pushed some documentation up as well. I haven't managed to test this as I don't have sphinx setup.

If you are all happy, then I squash!

rejeep:
Great, I don't have Sphinx set up either. Currently switching computers. Will check it out soon and then merge!

rejeep:
The documentation is not quite correct.

```diff
diff --git a/doc/guide/usage.rst b/doc/guide/usage.rst
index e21ea65..ff383c9 100644
--- a/doc/guide/usage.rst
+++ b/doc/guide/usage.rst
@@ -76,9 +76,10 @@ proper `$PATH` (see :ref:`cask path`) and `$EMACSLOADPATH` (see :ref:`cask
 load-path`).
 
 
-.. __cask:
+.. _cask eval:
 
 cask eval
+---------
 
 .. program:: cask eval
 
```

phillord:
I have pre-squashed commits then! If the it doesn't look fine, I fix and
squash again.

Johan Andersson <notifications@github.com> writes:

> Great, I don't have Sphinx set up either. Currently switching computers. Will check it out soon and then merge!
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/cask/cask/pull/296#issuecomment-106187038

-- 
Phillip Lord,                           Phone: +44 (0) 191 208 7827
Lecturer in Bioinformatics,             Email: phillip.lord@newcastle.ac.uk
School of Computing Science,            http://homepages.cs.ncl.ac.uk/phillip.lord
Room 914 Claremont Tower,               skype: russet_apples
Newcastle University,                   twitter: phillord
NE1 7RU

phillord:
Oh dear, email overlap. Resquashed.

Johan Andersson <notifications@github.com> writes:

> The documentation is not quite correct.
>
> ```diff
> diff --git a/doc/guide/usage.rst b/doc/guide/usage.rst
> index e21ea65..ff383c9 100644
> --- a/doc/guide/usage.rst
> +++ b/doc/guide/usage.rst
> @@ -76,9 +76,10 @@ proper `$PATH` (see :ref:`cask path`) and `$EMACSLOADPATH` (see :ref:`cask
>  load-path`).
>  
>  
> -.. __cask:
> +.. _cask eval:
>  
>  cask eval
> +---------
>  
>  .. program:: cask eval
>  
> ```
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/cask/cask/pull/296#issuecomment-106277684

-- 
Phillip Lord,                           Phone: +44 (0) 191 208 7827
Lecturer in Bioinformatics,             Email: phillip.lord@newcastle.ac.uk
School of Computing Science,            http://homepages.cs.ncl.ac.uk/phillip.lord
Room 914 Claremont Tower,               skype: russet_apples
Newcastle University,                   twitter: phillord
NE1 7RU

rejeep:
Awesome, thanks, merged! :+1:

rejeep:
@lunaryorn Do you have to release a new version for the docs to update?

lunaryorn:
@rejeep Documentation should update automatically on ReadTheDocs… unless we have a build failure, that is.  Can you check?  I'm on mobile currently…

rejeep:
@lunaryorn it took a while but now it's up to date, thanks!

Issue #285, opened 2015-03-15
Windows Install Error: TypeError: str expected, not bytes


sshaw:
Similar to #140.

Here's the output:

```
C:\Users\sshaw> curl -kfsSL https://raw.githubusercontent.com/cask/cask/master/go | python
Cloning into 'C:\Users\sshaw\.cask'...
remote: Counting objects: 3235, done.
remote: Compressing objects: 100% (5/5), done.
rRemote: Total 3235 (delta 0), reused 0 (delta 0), pack-reused 3230eceiving objects: 100% (3235/3235
Receiving objects: 100% (3235/3235), 1001.62 KiB | 1.29 MiB/s, done.

Resolving deltas: 100% (1812/1812), done.
Checking connectivity... done.
Traceback (most recent call last):
  File "C:\Users\sshaw\.cask\bin\cask", line 347, in <module>
    main()
  File "C:\Users\sshaw\.cask\bin\cask", line 330, in main
    exec_cask(sys.argv[1:])
  File "C:\Users\sshaw\.cask\bin\cask", line 292, in exec_cask
    emacs = get_cask_emacs()
  File "C:\Users\sshaw\.cask\bin\cask", line 257, in get_cask_emacs
    emacs = ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
  File "C:\Python34\lib\_collections_abc.py", line 428, in __contains__
    self[key]
  File "C:\Python34\lib\os.py", line 630, in __getitem__
    value = self._data[self.encodekey(key)]
  File "C:\Python34\lib\os.py", line 697, in encodekey
    return encode(key).upper()
  File "C:\Python34\lib\os.py", line 692, in check_str
    raise TypeError("str expected, not %s" % type(value).__name__)
TypeError: str expected, not bytes
←[31mCask could not be bootstrapped. Try again later, or report an issue at https://github.com/cask/
cask/issues←[0m
```

Also note that if I run `curl` + `go` again:

```
C:\Users\sshaw> curl -kfsSL https://raw.githubusercontent.com/cask/cask/master/go | python
←[31mDirectory C:\Users\sshaw\.cask exists. Is Cask already installed?←[0m
```

```
C:\Users\sshaw>python --version
Python 3.4.3
```

Let me know if you need more details

melihovv:
I have the same error. Pls help anybody

sshaw:
@melihovv the workaround is to run it with Python 2.

PR #297, opened 2015-06-16
Fix for issue #285 PY3 and Windows: TypeError exception

Ecno92:
Fix for issue #285 PY3 and Windows: TypeError when checking if INSIDE_EMACS is in ENVB

With this fix I was able to successfully install Cask and install my dependencies under Windows.
If you have the same issue please test this and react on this PR since I was the only one that tested it.

The changes are not as nice as I would normally make it, but at this moment I have no inspiration for a good refactor of the change which solves the problem and is compatible with both PY2 and PY3. Suggestions are welcome.

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -254,7 +254,10 @@ def get_cask_emacs():
     # compile.  In this case, $EMACS may be set as well to a rather
     # meaningless value.  If $INSIDE_EMACS is set, we just ignore the
     # value of $EMACS.
-    emacs = ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
+    if sys.version_info[0] == 2:
+        emacs = ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
+    else:
+        emacs = ENVB.get('EMACS') if 'INSIDE_EMACS' not in ENVB else None
     emacs = emacs or find_best_emacs()
     ensure_supported_emacs(emacs)
     return emacs


lunaryorn:
:-1:  That's not how this should be fixed.  `ENVB` is a *byte*-environment.  If Windows doesn't provide that, we should rather refactor the whole environment handling in the cask launcher.

Ecno92:
OK, thanks for the feedback. 
I'll look into it in the near future to provide a more robust solution and update the PR.

Ecno92:
I just did some checks since I'm not very experienced in working with environment variables yet. 
The problem is that Python 3 is more strict on the type which you use for the get on the ``os.environ``/``os.environb`` mapping. In Python 2 the situation as we can see in the current master's source code just works. (e.g. .get(b'EMACS')). On Python 3 it throws a error. In this case `os.environb` should be used since this returns a byte string mapping. Unfortunately `os.environb` is not available on Windows (https://docs.python.org/3.5/library/os.html#os.supports_bytes_environ). 

On my machines (Ubuntu/Windows) I can't see the advantage of using byte based environment variables (`os.environb`) in Python for this use case. Is this still something that should be maintained or is it OK to move to `os.environ` so we can move away from the byte strings? 

I thought it would be nice to first get some more insights before updating the pull request. I'm looking forward to your answer.

lunaryorn:
@Ecno92 I chose to use the byte environment for best compatibility with Python 2 and 3, since `os.environ` is a byte environment in Python 2 on Unix, and we didn't support Windows back then.

But I see that this was a wrong choice, and we should probably instead go for the string environment.  We need to make sure, though, that we properly encode/decode on Python 2 then.

If you'd like to refactor this, I'd be happy to review a pull request.  Note, though, that it _must_ work flawlessly on Python 2, since that is still the default Python on many systems, and we cannot rely on Python 3 being present on all systems.

Ecno92:
@lunaryorn I would like to fix this and I will also ensure compatibility between Python versions.
I'm quite busy this week, but I try to make some time available in the near future since this will be a nice exercise. 

I'll let you know once I get my hands dirty on this.

Ecno92:
I'm quite busy with all kinds of projects. So it is hard for me to start on this now.
Because it may take a long time before I can start with this I'm closing this PR. 

Also because I think #298 should be fixed first in order to solve this issue in an efficient way without breaking the PY2 build.

PR #299, opened 2015-06-25
Fix package command argument handling

politza:
A command like

```
$ cask package dist
```

may fail to create a package, because `package-build--build-multi-file-package` may change the `default-directory` and the package does end up in some temporary directory.


rejeep:
Just a minor note, but can you please change `expand-file-name` to `f-expand`.

PR #309, opened 2015-08-31
Update usage docs for Homebrew installation

akshaysmurthy:
The cask.el file is found in `/usr/local/share/emacs/site-lisp/cask/cask.el`
and not `/usr/local/share/emacs/site-lisp/cask.el`. The directory `cask`
is missing in the path.


rejeep:
Thanks! :+1:

PR #310, opened 2015-09-06
Add Cask to auto-mode-alist

zonuexe:
Edit `Cask` file using `lisp-mode`

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -921,6 +921,9 @@ a directory specified by `cask-dist-path' in the BUNDLE path."
         (f-mkdir target-dir))
       (package-build-package name version patterns path target-dir))))
 
+;; Edit `Cask' file using lisp-mode
+(add-to-list 'auto-mode-alist '("/Cask\\'" . lisp-mode))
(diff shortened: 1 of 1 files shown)


lunaryorn:
@zonuexe This doesn't belong into `cask.el`.  It's a non-interactive library, and generally Cask is a non-interactive tool.  Closing as wontfix.

rejeep:
@zonuexe You could instead update the documentation with a recommendation on adding that line to the Emacs configuration. I would recommend adding it on the [Cask DSL page](http://cask.readthedocs.org/en/latest/guide/dsl.html).

zonuexe:
@rejeep It is a good idea.  But, casual users may not read the document...

I know that Cask is not an interactive application. I also know that it is loaded from  batch-mode.
This line maybe don't interfere with Cask.  And, it will help to casual users.

rejeep:
Well, casual users will not add `cask` to their Emacs configuration, which would be necessary for this to load.

Issue #323, opened 2015-11-20
Use HTTPS to communicate with repositories where possible


hrs:
According to `cask-source-mapping`, Cask seems to connect to repositories over plain HTTP. This leaves us open to MITM attacks and a bunch of other problems.

From what I can tell, the `gnu`, `melpa`, `melpa-stable`, and `marmalade` sources currently support HTTPS. `SC` and `org` don't, yet.

I don't know very much about how Emacs handles TLS connections. Do we need to worry about certificates, or would a fix be as simple as changing "http" to "https" in `cask-source-mapping` for the sources that support it?

lunaryorn:
I don't know either.  On Unix systems Emacs should be able to create TLS connections, but I don't know whether Emacs actually checks certificates.   I know that Emacs 25 on OS X does that if it's build with GNU TLS, but that's about all that I know, and honestly the issue is not so important to me that I'd like to spend any time on researching that.

rejeep:
I have no idea either. @hrs if you are up for it you could dig into this and get back to us with your findings.

hrs:
Haha, @lunaryorn, fair enough.

And sure, @rejeep, I just looked into this a bit. Some folks have already looked into this issue:

* The folks at Melpa had a [big conversation](https://github.com/milkypostman/melpa/issues/2109) last year when they set up HTTPS.
* Glyph wrote a pretty good [blog post](https://glyph.twistedmatrix.com/2015/11/editor-malware.html) pointing out some more reasons not to use HTTP for package management in Emacs.

It looks like it wouldn't do any harm to switch to HTTPS addresses. `package.el` seems to fall back to HTTP if Emacs isn't configured for TLS, so I don't see a strong reason *not* to change.

If we *do* use HTTPS addresses, users that have built Emacs to handle TLS (with `gnutls` or `openssl`) will be able to securely use Cask without overwriting `cask-source-mapping` in their dotfiles.

I'm gonna go ahead and submit a PR changing the addresses that support TLS, if y'all don't mind.

mithrandi:
The HTTP fallback here renders the protection of HTTPS almost worthless; all an attacker has to do is make the HTTPS connection fail any way they like, and then they have an HTTP connection to toy with as they like.

lunaryorn:
@mithrandi I'm not sure what to make of your comment.  Where do we have a HTTP fallback here?

mithrandi:
[...]

PR #324, opened 2015-11-20
Use HTTPS to connect to sources that support it

hrs:
HTTP sends plaintext messages, so it leaks information and is vulnerable to MITM attacks. ELPA, MELPA, and Marmalade all support HTTPS. We should make use of it.

This changes the default addresses for those sources to use HTTPS. Users that have built Emacs to handle TLS will be able to make use of these sources, and Emacs will fall back to using regular HTTP if it isn't configured to handle TLS.

This also updates the documentation to point to the HTTPS address of those repositories.

This should close https://github.com/cask/cask/issues/323.

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -52,8 +52,8 @@
         (condition-case nil
             (mapc 'require cask-bootstrap-packages)
           (error
-           (add-to-list 'package-archives (cons "gnu" "http://elpa.gnu.org/packages/"))
-           (add-to-list 'package-archives (cons "melpa" "http://melpa.org/packages/"))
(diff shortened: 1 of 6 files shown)


lunaryorn:
LGTM 👍. Thank you for making us aware, and for fixing the problem!  👏👍

lunaryorn:
@rejeep Shall we merge this?

rejeep:
I haven'd had time to try it out yet. You make the call! :+1:

lunaryorn:
I'll just merge then and look how it goes 😎

rejeep:
:boom:

rejeep:
:smile:

Issue #321, opened 2015-11-12
Feature Discussion: Plugin System


AdrieanKhisbe:
Few months ago in order to automate my emacs-packages I end up building [cask-package-toolset](https://github.com/AdrieanKhisbe/cask-package-toolset.el).

It's a tool to help setup continous integration for an emacs project, relying on cask, evm, ert, ecukes, undercover, travis and some other cool stuff.

So far, it has to be added as a development dependancy to `Cask`, installed and be invoked with `cask exec package-toolset status`. 
_As this is not the best developper experience, I was considering it could be a good idea to integrate as a *plugin to cask_, and so to build **an extensible plugin system** for Cask so it becomes easy to  to add arbitrary commands to bundler itself (like your bundle license command)*
(similar way that the heroku or [bundler allow plugins](https://github.com/bundler/bundler/issues/1945))

I suggested the idea to @rejeep, that said it was a rather food idea that should be discussed on the repo.

_So what you think about this? =)_
Any suggestion about the feature, and it's possible implementation is very much welcome!!

lunaryorn:
@AdrieanKhisbe I like the idea, but it's a major task to implement that, and I for my part have no time for this.

rejeep:
Yes, this is a good idea, but like @lunaryorn says, quite a major change and my time is very limited as well.

I would suggest as a first step to write a spec for exactly what should be possible to extend. Write a list of use cases; how it works now and how a plugin system would solve it.

AdrieanKhisbe:
Indeed that's a major change, I won't deny :)

I'll try to draft something in the days to come.

conao3:
Sorry, `cask` is now shrinking state and this kind of feature request is heavy.
Closing for now, but any contribution on it is welcome anytime.

PR #326, opened 2015-12-12
Ci and package toolset

AdrieanKhisbe:
Improved CI and README:

have a look here: 
- README: https://github.com/AdrieanKhisbe/cask/tree/ci-and-package-toolset
- build: https://travis-ci.org/AdrieanKhisbe/cask/builds
- coverage: https://coveralls.io/github/AdrieanKhisbe/cask

What PR consist:
- Update of travis with evm 0.7 (-> containers, evm-travis)
- coverage for unit test. 
- ci badge to README
-  some more introduction to README
- scaffolded `Cask`  file now countain [cask-package-toolset](https://github.com/AdrieanKhisbe/cask-package-toolset.el), tool to help set up CI on cask project.

What could be idealy done:
add coverage for features test, but that would require to ack around (cf https://github.com/AdrieanKhisbe/cask-package-toolset.el/blob/master/features/step-definitions/cask-package-toolset-steps.el#L7-L28)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,9 @@ env:
   - EVM_EMACS=emacs-24.4-travis
   - EVM_EMACS=emacs-24.5-travis
 script:
+  - emacs --version
   - make test
+
 after_failure:
   - cat test/servant/tmp/servant.log
 after_script:
--- a/Cask
+++ b/Cask
@@ -16,4 +16,6 @@
  (depends-on "el-mock")
  (depends-on "noflet")
  (depends-on "ert-async")
- (depends-on "shell-split-string"))
+ (depends-on "shell-split-string")
+ (depends-on "cask-package-toolset")
+ (depends-on "undercover"))
--- a/README.md
+++ b/README.md
@@ -2,6 +2,13 @@
 
 ![Cask](cask.png)
 
+[![Build Status](https://travis-ci.org/cask/cask.svg)](https://travis-ci.org/cask/cask)
+[![Coverage Status](https://coveralls.io/repos/cask/cask/badge.svg)](https://coveralls.io/r/cask/cask)
+[![MELPA](http://melpa.org/packages/cask-badge.svg)](http://melpa.org/#/cask)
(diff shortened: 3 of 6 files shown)


AdrieanKhisbe:
(quick fix, forgot I already forked, hence worked on a non up to date repo.
Then I pull and rebase my modification, hence the push-f)

lunaryorn:
Thank you for this elaborate changeset.  I like your changes to the README, but I'm unsure about coveralls and about cask-package-toolset.

While I'm all for the idea of measuring test coverage, I'd like to know how it measures test coverage, and how much work it'd be to integrate the feature tests which are somewhat important to us.

As for cask-package-toolset, what is it, and how does it us and our users?  I'm careful to recommend 3rd party projects in our documentation… and if it's important or useful, shouldn't we make it part of Cask itself?

AdrieanKhisbe:
Hey @lunaryorn! :)

#### About coverage,
it is performed with [undercover](https://github.com/sviridov/undercover.el)

 To add them to the feature test, two option:
   - either the step call the commander command via the function in the same process
   - either we make run the cli instrumented

 On cpt, I went for the first option, but it does not handle signals. (and cask test rely on them 

### About [Cask package Toolset]()

*Cask Package Toolset*, *aka* cpt, is a cli tool to help setup test, coverage and other emacs package best practice. (badges, melpa recipe)

Ideally it would be totally integrated to cask in a plugin system, https://github.com/cask/cask/issues/321
but would take time to implement, I went for the pragmatic way to add it by default to `cask init`

I have no problem to migrate the repo it under *Cask* organisation if you accept me in it :)

rejeep:
> how much work it'd be to integrate the feature tests

Agreed. The reason the test coverage is not 100% is because some parts are tested using the feature tests.

Would it be possible to use [Ecukes before hooks](https://github.com/ecukes/ecukes#before) for this?

AdrieanKhisbe:
Not sure, the step itself has a call to an external process and check the output:

```lisp
(apply
               'call-process
               (append (list cask-test/bin-command nil (current-buffer) nil) args))))
```

conao3:
Thanks, but we had moved GitHub Actions, so this PR is not needed for now...? Thanks again for your contribution!

Issue #330, opened 2016-01-08
Running inside emacs


cpitclaudel:
This piece of code is causing me headaches :)

```
    # If $INSIDE_EMACS is set, Cask is called from within Emacs, e.g. M-x
    # compile.  In this case, $EMACS may be set as well to a rather
    # meaningless value.  If $INSIDE_EMACS is set, we just ignore the
    # value of $EMACS.
    emacs = ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
    emacs = emacs or find_best_emacs()
    ensure_supported_emacs(emacs)
    return emacs
```
- Running `M-x compile RET EMACS=emacs24 cask build` compiles the current project for emacs25
- Running from the emacs shell (either `shell` or `eshell`) doesn't allow one to specify which version of emacs to use.

This forces me to use the regular command line, but then I can't jump to errors directly from the compile buffer. Is there a workaround?

lunaryorn:
@cpitclaudel Not that I know of.  The situation around `$EMACS` is pretty unfortunate; it's almost entirely useless inside Emacs thus we can't really rely on its value, and we have no chance to figure out whether you explicitly overrode it, as in your `M-x compile` invocation.

We'd need to add an alternative `CASK_EMACS` variable probably.  Personally I never invoke Cask from inside Emacs, though :)

phillord:
It would be good to get this fixed in Emacs. The "INSIDE_EMACS" behaviour is new in Emacs-25, so it's possible to change. Would something like "EMACS=~/path-to-current/emacs" make sense?

`M-x compile RET EMACS=emacs24 cask` should still work since it would reset the default emacs value.

At the moment, I am running cask of Emacs-25 trunk and cask is running Emacs-24.

PR #335, opened 2016-02-23
Add support for CASK_EMACS environment variable.

phillord:
See also #330

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -236,26 +236,48 @@ def find_best_emacs():
     else:
         return 'emacs'
 
+def inside_emacs_24():
+    """Whether Cask is running inside Emacs-24.
+
+    Emacs 24 is rather inconsistent in its use of the ``$EMACS`` and
+    ``$INSIDE_EMACS`` environment variables, so we try here to test if Cask is
+    being run inside Emacs (either in a shell or with ``M-x compile``).
+
+    Return ``True`` when running in Emacs 24, ``False`` otherwise.
+
+    """
+    ## are we inside Emacs at all
+    return (b'INSIDE_EMACS' in ENVB and
+            ## M-x compile in Emacs-24 sets INSIDE_EMACS to exactly 't'
+            (ENVB.get(b'INSIDE_EMACS') == 't' or
+             ## while M-x shell sets it to a string, starting with the
+             ## version number
+             ENVB.get(b'INSIDE_EMACS').startswith('24')))
+
+def get_emacs_from_env():
+    """Get the Emacs executable as specified by the environment.
+
+    Use the command given by the environment variable ``$CASK_EMACS`` if set.
+    Failing that use ``$EMACS``, unless the value of this variable does not
+    refer to a real Emacs executable (i.e. Cask is run from inside Emacs).
+
+    """
+    return (ENVB.get(b'CASK_EMACS') or
+            (ENVB.get(b'EMACS') if not inside_emacs_24() else None))
 
 def get_cask_emacs():
     """Get the Emacs executable to use for Cask.
 
-    Try to use the command denoted by the environment variable ``$EMACS``.  If
-    the value of this variable does not refer to a real Emacs executable
-    (i.e. Cask is run from inside Emacs), automatically find a good Emacs for
-    the current platform (see ``find_best_emacs``).
+    If an Emacs is defined by the environment (see ``get_emacs_from_env``) use
+    that, or automatically find a good Emacs for the current platform (see
+    ``find_best_emacs``).
 
     Return the name or path for the Emacs command Cask shall use as string.
     Raise ``UnsupportedEmacsVersionError``, if the Emacs command does not meet
     the minimum version requirements of Cask.
 
     """
-    # If $INSIDE_EMACS is set, Cask is called from within Emacs, e.g. M-x
-    # compile.  In this case, $EMACS may be set as well to a rather
-    # meaningless value.  If $INSIDE_EMACS is set, we just ignore the
-    # value of $EMACS.
-    emacs = ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
-    emacs = emacs or find_best_emacs()
+    emacs = get_emacs_from_env() or find_best_emacs()
     ensure_supported_emacs(emacs)
     return emacs
 
@@ -275,6 +297,12 @@ def exec_command(command):
     # Copy the environment and update the paths
     ENVB[b'EMACSLOADPATH'] = get_cask_path('load-path')
     ENVB[b'PATH'] = get_cask_path('path')
+
+    ## special handling for where we have a CASK_EMACS, we can use that and
+    ## pass it as the EMACS variable to the command
+    if (b'CASK_EMACS' in ENVB):
+        ENVB[b'EMACS'] = ENVB[b'CASK_EMACS']
+
     os.execvp(command[0], command)
 
 
--- a/doc/guide/usage.rst
+++ b/doc/guide/usage.rst
@@ -52,6 +52,13 @@ the `emacs` command.  To pick a different Emacs, set the environment variable
 
    $ EMACS="emacs24.1" cask command
 
+Exceptionally, if you are launching :program:`cask` inside Emacs 24 either
+from an internal shell or `M-x compile`, then Emacs uses :envvar:`EMACS` in a
+way which conflicts with :program:`cask`, in which case you can use the
+environment variable :envvar:`CASK_EMACS` instead. With Emacs 25,
+:envvar:`EMACS` can be used as normal.
+
+
 Commands and options
 ====================
 
@@ -440,3 +447,8 @@ Environment variables
    - ``emacs``
 
    On other Unix variants, e.g. Linux, Cask will simply use ``emacs``.
+
+.. envvar:: CASK_EMACS
+
+   As EMACS, but takes precedence over it. This is most useful for launching
+   Cask inside Emacs which often resets EMACS to other values.


bin/cask
```diff
@@ -254,7 +254,9 @@ def get_cask_emacs():
     # compile.  In this case, $EMACS may be set as well to a rather
     # meaningless value.  If $INSIDE_EMACS is set, we just ignore the
     # value of $EMACS.
+    #print( ENVB )
```
lunaryorn: Please remove debugging print statements.

bin/cask
```diff
@@ -275,6 +277,8 @@ def exec_command(command):
     # Copy the environment and update the paths
     ENVB[b'EMACSLOADPATH'] = get_cask_path('load-path')
     ENVB[b'PATH'] = get_cask_path('path')
+    # Set EMACS up for tools which use this
+    ENVB[b'EMACS'] = get_cask_emacs()
```
lunaryorn: I'm not sure about this change, please see my comment on the pull request.

bin/cask
```diff
@@ -254,7 +254,9 @@ def get_cask_emacs():
     # compile.  In this case, $EMACS may be set as well to a rather
     # meaningless value.  If $INSIDE_EMACS is set, we just ignore the
     # value of $EMACS.
+    #print( ENVB )
     emacs = ENVB.get(b'EMACS') if b'INSIDE_EMACS' not in ENVB else None
+    emacs = ENVB.get(b'CASK_EMACS') or emacs
```
lunaryorn: Please change this code to check this variable first, before the other one.  It's strange to check a variable that's supposed to take precedence at last.  

Please also change this code to not require three assignments to `emacs`.  I guess you can move the environment variable access into a separate function, so that the line below just reads `emacs = get_emacs_from_env() or find_best_emacs()`.

lunaryorn:
@phillord Thank you very much for this change.

I like the idea of `$CASK_EMACS`, but resetting `$EMACS` for subprocesses of Cask is a totally different story.  That's a breaking change for sure, and while I believe that it would not adversely affect existing uses of Cask, this behaviour—namely that Cask forcibly exports the path to it's own host Emacs in `$EMACS`—should at the very least be documented explicitly, and it should be discussed separately and not be sneaked into an unrelated pull request.

We can perfectly export the host Emacs in `$CASK_EMACS`, i.e. forcibly set `$CASK_EMACS` for subprocesses once we found a proper Emacs, but we should not carelessly change a variable that Emacs itself uses.

Please remove this part from the pull request, and open a separate one to discuss it independently.  Thank you :relaxed:

phillord:
Other changes you have requested, yes, will address.

The reason for the $EMACS change is that ert-runner doesn't work otherwise (because it looks for $EMACS). If I understand the percolation of environment variables though pythons process handling correctly, if $EMACS were set then it would percolate down to ert-runner, but as it isn't, obviously, it doesn't.
$CASK_EMACS does percolate down but, of course, ert-runner ignores that. All of which tells me this problem is an issue for this PR.

Solutions I can see:

1) Set $EMACS as per current PR, but ONLY to the value of $CASK_EMACS. This is non-breaking as there is no current CASK_EMACS handling.
2) Make ert-runner handle CASK_EMACS also. This is okay, but only works for ert-runner.
3) What I did. I think this is right, myself, since if exec runs emacs (as with ert-runner), what I want, what I really, really want is to be to say which emacs (zigazig ha).

phillord:
I've updated with option 1 from above. Without this, having CASK_EMACS doesn't help me, because ert-runner fails.

bin/cask
```diff
@@ -237,25 +237,30 @@ def find_best_emacs():
...
 
+def get_emacs_from_env():
+    """Get the Emacs executable as specified by the environment.
+
+    Use the command given by the environment variable ``$CASK_EMACS`` if set.
+    Failing that use ``$EMACS``, unless the value of this variable does not
+    refer to a real Emacs executable (i.e. Cask is run from inside Emacs).
+
+    """
+    return ENVB.get(b'CASK_EMACS') or \
```
lunaryorn: Please use parenthesis around the whole expression.  Breaking lines with backslashes is discouraged in Python.

lunaryorn:
@phillord I would prefer if we could update ert-runner to use `$CASK_EMACS` instead.

`$EMACS` is simply not reliable, because Emacs overwrites it at some points.  For instance, `M-x compile` in `cask exec emacs` would still suffer from the same issue.

phillord:
@lunaryorn As far as I can tell, what you say about $EMACS is not true for a build from emacs-25. It does not set up $EMACS at all. Although it does set $INSIDE_EMACS up. This means that both Cask and ert-runner will still ignore a correctly set $EMACS. So, while I understand the reason for your heuristic, I think it is wrong.

The $CASK_EMACS solution seems sub-optimal to me because it introduces extra handling.

Another suggestion (for both cask and ert-runner) it to modify the $INSIDE_EMACS operation so that this only happens when $INSIDE_EMACS starts with "24". I am happy to do this for both cask and ert-runner. But I need some solution because, as it stands, this largely prevents me from using cask effectively.

lunaryorn:
@phillord Emacs 24 sets `$EMACS`; that's were all this hassle comes from.  And it doesn't even set `$INSIDE_EMACS` consistently as far as I know.

> The $CASK_EMACS solution seems sub-optimal to me because it introduces extra handling.

What "handling" specifically?  Where's the difference between setting `$EMACS` and `$CASK_EMACS`?

In any case, `$CASK_EMACS` is preferable because it's *our* variable, i.e. only Cask knows about it and only Cask ever changes it.  The less conflicts the better.

> Another suggestion (for both cask and ert-runner) it to modify the $INSIDE_EMACS operation so that this only happens when $INSIDE_EMACS starts with "24". I am happy to do this for both cask and ert-runner. 

No offense meant, but that's even worse than what Cask does currently, because the heuristic is even more complex and fragile.

phillord:
Sebastian Wiesner <notifications@github.com> writes:

> @phillord Emacs 24 sets `$EMACS`; that's were all this hassle comes from. And
> it doesn't even set `$INSIDE_EMACS` consistently as far as I know.
>
>> The $CASK_EMACS solution seems sub-optimal to me because it introduces extra handling.
>
> What "handling" specifically? Where's the difference between setting `$EMACS`
> and `$CASK_EMACS`?

Well, cask has to know about CASK_EMACS as well as EMACS. As do other tools.

> In any case, `$CASK_EMACS` is preferable because it's *our* variable, i.e.
> only Cask knows about it and only Cask ever changes it. The less conflicts the
> better.

Yep. But ert-runner has to handle it also, otherwise, it fails of
purpose. Also ecukes. Also, what ever other downstream tools we might
wish to exec.

>> Another suggestion (for both cask and ert-runner) it to modify the
>> $INSIDE_EMACS operation so that this only happens when $INSIDE_EMACS starts
>> with "24". I am happy to do this for both cask and ert-runner.
>
> No offense meant, but that's even worse than what Cask does currently, because
> the heuristic is even more complex and fragile.

No offense taken at all. The heuristic is ugly, but it has some
advantages over the current situation.

It's actually a refinement of the current heuristic (albeit a more
complex one). It works in Emacs-25 which the current heuristic does not,
I think (in fact, it breaks the simpler behaviour that should work).
And, finally, it's a temporary heuristic. It can be removed after a
while once emacs-25 is out.

Your call. If you don't want this fix, I'll complete the PR for
CASK_EMACS and send another in for ert-runner.

Phil

lunaryorn:
@phillord Surely downstream tools need to know about `$CASK_EMACS`.  But currently they need to know about `$EMACS`, too—this variable doesn't appear from nowhere either.  Where's the difference?  Except that we can safely make promises about the former, but—due to Emacs 24' unfortunate behaviour—not about the latter.

The situation around `$EMACS` is unfortunate, but any attempt to add even more complex heuristics for `$EMACS` isn't going to get us anywhere.  We should strive for _simpler_ code, all the more since we can't just remove code for Emacs 24 any time soon, not even after Emacs 25 comes out.  Cask is a development tool, as such we have to support all Emacsen that extensions target, and are not in a position to just go ahead and redefine the version baseline for everyone else.  We're bound to support Emacs 24 for a long time.

lunaryorn:
> Your call. If you don't want this fix, I'll complete the PR for CASK_EMACS and send another in for ert-runner.

If you're willing to do that, I'd prefer this approach, to be honest.

That's a lot easier to document and to understand for all users, because it comes down to

- Cask reads its Emacs from `$CASK_EMACS`, and applies heuristics to `$EMACS` only as a fallback.
- Cask exposes its host Emacs in `$CASK_EMACS`, where all tools can easily take a proper and valid path to the host emacs.

However, I'd like to wait for @rejeep to join our discussion and give his ok.

phillord:
Sebastian Wiesner <notifications@github.com> writes:

> @phillord Surely downstream tools need to know about `$CASK_EMACS`. But
> currently they need to know about `$EMACS`, too—this variable doesn't appear
> from nowhere either. Where's the difference?

Downstream tools need to know about BOTH $CASK_EMACS and $EMACS.
Otherwise, they will fail if $EMACS is used outside of emacs. Since
$EMACS is pretty standard, I think this is introducing more complexity.

> Except that we can safely make promises about the former, but—due to
> Emacs 24' unfortunate behaviour—not about the latter.

I agree. Since Emacs-24 blitzes $EMACS passed from outside, there is no
way around this.


> The situation around `$EMACS` is unfortunate, but any attempt to add even more
> complex heuristics for `$EMACS` isn't going to get us anywhere.

It is certainly true that using a heuristic which only operates on
emacs-24 has longer code, and in the sense is more complex. But, it also
makes the heuristic more limited in scope, and that is surely a good
thing. As we can see, in Emacs-25 actually breaks things that should
work straight-forwardly.

> We should strive for _simpler_ code, all the more since we can't just
> remove code for Emacs 24 any time soon, not even after Emacs 25 comes
> out. Cask is a development tool, as such we have to support all
> Emacsen that extensions target, and are not in a position to just go
> ahead and redefine the version baseline for everyone else. We're bound
> to support Emacs 24 for a long time.

But pushing the CASK_EMACS var to downstream tools is, I think, more
complex. And more complex for the user.

Here I think is a solution:

1) Add a CASK_EMACS variable as per PR
2) Check INSIDE_EMACS starts 24
   If so, use CASK_EMACS, reset EMACS=CASK_EMACS before exec
   
3) Also set INSIDE_EMACS to nil, or remove INSIDE_EMACS handling from
   ert-runner

From user perspective we now have

1) For Emacs-25, or outside emacs use $EMACS
2) For Emacs-24, use $CASK_EMACS
3) Know that $CASK_EMACS handling is deprecated, and is there for
Emacs-24 wierdness.


This means that special case handling is very limited for the user, only
to those cases where it is actually necessary.

lunaryorn:
Why do you assume that `$EMACS` the way we use it is some kind of a "standard"?  On the contrary, for the longest time `$EMACS` meant `$INSIDE_EMACS`; no tool could rely on `$EMACS` pointing to an Emacs executable.  

As far as ert-runner and all other Cask tools, we _defined_ this "standard".  ert-runner doesn't look at `$EMACS` because it's a standard.   It looks there because that is what Cask does, and for no other reason.

If Cask exported `$CASK_EMACS` ert-runner would just look there; it's meant and built to be used from within Cask, and can safely rely on Cask-specific variables.  You're not running `.cask/ert-runner-<version>/bin/ert-runner`, are you?

I'm not sure where you see complexity for the user… what's the difference between setting `$EMACS` and `$CASK_EMACS`?   

That said I wonder about your perception of "complexity".  You're arguing against my position on the grounds of complexity, but then use more text, more bullet points, more "ifs" and "on emacs version x.y" to propose your supposedly simpler approach?  The irony in that seems to have escaped you.

My proposal is:  Expect `$CASK_EMACS`, export `$CASK_EMACS`, and let downstream tools of Cask use `$CASK_EMACS`, and `$CASK_EMACS`  *alone*.  IOW, replace `$EMACS` with `$CASK_EMACS` and get rid of all the mess right away.  How can you get simpler than that?

lunaryorn:
And besides, Cask doesn't look at `$EMACS` because it's some kind of a standard.  Cask looks at `$EMACS` because we thought that it'd be a nice variable to tell Cask about the path to Emacs, before we discovered that `$EMACS` has strange values inside a running Emacs, and before we learned that a "namespaced" variable like `$CASK_EMACS` is safer.

Had we used `$CASK_EMACS` from start you'd not even realise that there was `$EMACS` as well.

rejeep:
First of all, @phillord it would have been nice with a pull request description. I don't want to read the code to understand the purpose. In this case, I don't even know Python. Sure I can figure it out. But this is not only for me, it's for everyone out there using Cask and want to follow along in the development. Some might not even know any Emacs Lisp.

Using the variable `$EMACS` (that exact name) was a mistake. The purpose of it was to be able to run Cask with any Emacs without fiddling with binaries and aliases, etc. The purpose is _"Run Cask using this Emacs"_. I did the same mistake in Ecukes and Ert runner.

The variable should have been called `$CASK_EMACS`, `$ERT_RUNNER_EMACS`, `$ECUKES_EMACS`, etc...

And sure, we could add a `$CASK_EMACS` variable now, but it would take a long time before we could deprecate using `$EMACS`, so I don't really see the point, but I'm not against it.

> But I need some solution because, as it stands, this largely prevents me from using cask effectively.

Here's were I would like to see a good pull request description. Why cant you use Cask without this? For example, why cant you simply set `$EMACS` yourself?

> Yep. But ert-runner has to handle it also, otherwise, it fails of purpose. Also ecukes. Also, what ever other downstream tools we might wish to exec.

I agree that tools such as Ecukes and Ert runner should not depend or know anything of Cask. You should be able to run them without Cask.

phillord:
Sebastian Wiesner <notifications@github.com> writes:
> Why do you assume that `$EMACS` the way we use it is some kind of a
> "standard"? On the contrary, for the longest time `$EMACS` meant
> `$INSIDE_EMACS`; no tool could rely on `$EMACS` pointing to an Emacs
> executable.

So, I checked and you are correct. I was getting confused somewhere
between $EDITOR (the environment variable) and EMACS (the make file
variable).


> I'm not sure where you see complexity for the user… what's the difference
> between setting `$EMACS` and `$CASK_EMACS`?

$EMACS seems to be the obvious name, that's all. Nothing more than that.
I presume that's what it was chosen in the first place.


> That said I wonder about your perception of "complexity". You're arguing
> against my position on the grounds of complexity, but then use more text, more
> bullet points, more "ifs" and "on emacs version x.y" to propose your
> supposedly simpler approach? The irony in that seems to have escaped you.

I was trying to be precise, is all. $EMACS seems simpler, but is broken
when launching cask from inside Emacs 24 but not from Emacs 25. Solution
-- make a work around for 24 now, then drop in a years time.


> My proposal is: Expect `$CASK_EMACS`, export `$CASK_EMACS`, and let downstream
> tools of Cask use `$CASK_EMACS`, and `$CASK_EMACS` *alone*. IOW, replace
> `$EMACS` with `$CASK_EMACS` and get rid of all the mess right away. How can
> you get simpler than that?

It's fine. I think my alternative is nicer, and you think it's harder.
Okay, no worries, it's not my project, I'll do it any way you want.

Phil

phillord:
Johan Andersson <notifications@github.com> writes:

> First of all, @phillord it would have been nice with a pull request
> description. I don't want to read the code to understand the purpose. In this
> case, I don't even know Python. Sure I can figure it out. But this is not only
> for me, it's for everyone out there using Cask and want to follow along in the
> development. Some might not even know any Emacs Lisp.

Okay, the issue where this discussed was linked.


> Using the variable `$EMACS` (that exact name) was a mistake.

Seems like a reasonable name to me -- just unfortunate that it
contradicts it's usage in Emacs 24.

>> But I need some solution because, as it stands, this largely prevents me from using cask effectively.
>
> Here's were I would like to see a good pull request description. Why cant you
> use Cask without this? For example, why cant you simply set `$EMACS` yourself?

Because you unset it again, and I cannot prevent this.

I am running Emacs-25 (or master) in place from a git checkout.

I have a cask managed project which depends on Emacs-25. I am running
cask from a makefile, which looks like this.

    test:
	EMACS=$(EMACS) $(CASK) exec ert-runner

All works, everyone is happy. Except that I normally run make from
inside Emacs, using M-x compile. In this usage, it fails, because M-x
compile sets up $INSIDE_EMACS. Both cask and ert-runner see this, and as
a result ignore the (correct) value of $EMACS. Emacs is now run as
"emacs" which in my case means "/usr/bin/emacs" which is emacs-24.5.
And, so, my tests fail.

Conclusion: the heuristic that cask (and ert-runner) uses to set Emacs
is flawed for Emacs-25. I see no good workaround.


> Yep. But ert-runner has to handle it also, otherwise, it fails of purpose.
>> Also ecukes. Also, what ever other downstream tools we might wish to exec.
>
> I agree that tools such as Ecukes and Ert runner should not depend or know
> anything of Cask. You should be able to run them without Cask.


So, we have a number of solutions to the problem, given in this thread:
I shall not recap for fear of writing "more text, more bullet points and
more ifs".

I've written enough now, so if either you or @lunaryorn want to decide,
I volunteer to implement it both for cask and ert-runner.

rejeep:
@phillord Why don't you just unset `$INSIDE_EMACS`?

lunaryorn:
@rejeep That wouldn't work because many commands (e.g. `M-x compile`) set it automatically, without any chance to prevent it.

phillord:
Johan Andersson <notifications@github.com> writes:
> @phillord Why don't you just unset `$INSIDE_EMACS`?


It's possible, I guess. I could do "M-x compile" then

unset INSIDE_EMACS;make test

But ugly.

I'm afraid that the INSIDE_EMACS handling is just not doing the right
thing for Emacs-25. I'm afraid, it's not doing the right thing for
Emacs-24 either, but it may be doing the only thing it can there.

Phil

rejeep:
I see a lot of solutions here, but not a really good one. What if we add an `--emacs` option (that will set `$EMACS` and perhaps even `$CASK_EMACS`)? That way you could run:

```
$ cask exec something --emacs /path/to/emacs
```

We would have to handle the option parsing in Python, which is a bit unfortunate.

@phillord I'm not even sure this is a better solution. But would it solve the issue for you?

phillord:
@rejeep I started putting a --emacs option in a while back (see #295) but I never got it to work to completion. Yes I think this is a nice option, and it does not suffer from the fragility of the environment I think.

Will it solve the issue? Yes, iff we also add support to ert-runner as well. The problem at the moment is that cask is running on Emacs and ert-runner is running another. And this will make a nasty command line because I will have to give Emacs twice.

It's possible that the core problem here is the over all way of launching ert-runner. Currently I am using an interactive Emacs to launch python (cask) which runs Emacs (cask.el) which runs a shell script (ert-runner) which runs and Emacs which runs the tests (ert-runner.el). Perhaps the solution is to not use cask exec, but instead have cask load ert-runner.el in its own Emacs. Something like....

      cask eval "(progn (load-library "ert-runner.el") (ert-runner/run))"

Then, at least, we only have to get the environment right once

rejeep:
> It's possible that the core problem here is the over all way of launching ert-runner. Currently I am using an interactive Emacs to launch python (cask) which runs Emacs (cask.el) which runs a shell script (ert-runner) which runs and Emacs which runs the tests (ert-runner.el).

That's a nasty chain of events, no wonder it's hard to get right! :smile:

phillord:
Yep. It might be that ert-runner is not helping me a lot here. Unfortunately, a quick experiment at launching ert-runner directly fails because of the use of commander at the end -- it expects command line arguments, so crashes if directly loaded.

rejeep:
Yeah, that should really be split into a library part and a cli part, like we do for Cask.

phillord:
So, to recap where we are at the moment and what I would like to do.

1) I would like to add CASK_EMACS support still (because it is necessary for Emacs 24). 
2) I would like to limit the current INSIDE_EMACS behaviour to Emacs 24, since it is wrong for 25.
3) Split ert-runner up, so that it can be run with "cask eval" rather than "cask exec"
4) Likewise limit the INSIDE_EMACS behaviour to Emacs24 for ert-runner

Possibly, another command line option needs to be added to cask which, effectively, does eval, but looks a bit nicer than putting lisp on the command line. 

For my workflow (running cask inside Emacs 25, and testing other versions outside Emacs on travis) I need either 2 and 3, or  2 and 4.

I'm willing to do all of this, and I'd propose to do each in separate PRs. Comments?

lunaryorn:
I've given my opinion, which hasn't changed since.  I still consider this approach to be much more complicated than simply pushing `$CASK_EMACS` downstream, and I don't think that anyone would ever run `ert-runner` outside of Cask.  

But @rejeep obviously disagrees, as such I'll opt-out from this discussion and let you two sort things out :smile:

phillord:
Current state:

- Deleting EMACS now happens only inside Emacs-24
- CASK_EMACS is implemented and overrides everything
- If CASK_EMACS is set, it is passed through to exec as EMACS, and INSIDE_EMACS is deleted.

This allows:

- Outside Emacs, EMACS can be used (or CASK_EMACS)
- Inside Emacs-24, CASK_EMACS must be used
- Inside Emacs-25, EMACS can be used, except for ert-runner where CASK_EMACS is needed

The last caveat for Emacs-25, I can remove by patching ert-runner to handle $EMACS and $INSIDE_EMACS in its script, in the same way as I have done for Cask.

@lunaryorn I know it's more complicated, but it preserves the use of $EMACS except where Emacs deletes it.

rejeep:
> But @rejeep obviously disagrees, as such I'll opt-out from this discussion and let you two sort things out :smile:

As there is no really good solution to this, I'm willing to accept it.

phillord:
I think this PR is about ready now (bar squashing).

Just to recap (because it's been long):

 - Ignore EMACS only on Emacs 24
 - Accept CASK_EMACS, overrides EMACS
 - If CASK_EMACS is set remove INSIDE_EMACS from subprocesses.

rejeep:
@lunaryorn you will have to be the Python judge here

bin/cask
```diff
@@ -236,26 +236,45 @@ def find_best_emacs():
     else:
         return 'emacs'
 
+def inside_emacs_24():
+    """Return true if we are running inside Emacs-24.
```
lunaryorn: Use `Whether Cask is running inside Emacs 24`

bin/cask
```diff
@@ -236,26 +236,45 @@ def find_best_emacs():
...
+
+Emacs 24 is rather inconsistent in its use of the EMACS and INSIDE_EMACS
```
lunaryorn: `$EMACS` and `$INSIDE_EMACS`, please, and also use backticks to mark both as literals.

bin/cask
```diff
@@ -236,26 +236,45 @@ def find_best_emacs():
...
+environment variables, so we try here to test if Cask is being run inside
+Emacs (either in a shell or with M-x compile)"""
```
lunaryorn: Use backticks around `M-x compile`, and please end docstrings with a period.

lunaryorn: Also, add a final sentence in a new paragraph, saying:

```
Return ``True`` when running in Emacs 24, ``False`` otherwise
```

bin/cask
```diff
@@ -236,26 +236,45 @@ def find_best_emacs():
...
+    ## are we inside Emacs at all
+    return (b'INSIDE_EMACS' in ENVB and
+            ## M-x compile in Emacs-24 sets INSIDE_EMACS to exactly "t"
+            (ENVB.get(b'INSIDE_EMACS') == "t"
```
lunaryorn: Please use single quotes for strings.

bin/cask
```diff
@@ -236,26 +236,45 @@ def find_best_emacs():
...
+             or
+             ## while M-x shell sets it to a string, starting with the
+             ## version number
+             ENVB.get(b'INSIDE_EMACS').startswith("24")))
```
lunaryorn: Likewise.

bin/cask
```diff
@@ -275,6 +294,16 @@ def exec_command(command):
...
     ENVB[b'PATH'] = get_cask_path('path')
+
+    ## special handling for where we have a CASK_EMACS, we can use that and
+    ## pass it as the EMACS variable to the command
+    if (b'CASK_EMACS' in ENVB):
+        ENVB[b'EMACS'] = ENVB[b'CASK_EMACS']
+        ## we also remove INSIDE_EMACS because EMACS should now be
+        ## trustworthy.
+        if b'INSIDE_EMACS' in ENVB:
+            del ENVB[b'INSIDE_EMACS']
```
lunaryorn: I don't like this.  Even tools running within Emacs and Cask might have a valid reason to find out whether they run in Emacs.  We should preserve `$INSIDE_EMACS` under all circumstances, and I strongly disagree with dropping it here, no matter what the reason is.

phillord: Yes, it's a hack. The fix to ert-runner avoids the necessity for this now, although other tools (like ecukes) will need patching or they will fail.

lunaryorn:
LGTM now.

phillord:
All squashed.

rejeep:
@lunaryorn feel free to merge when you're happy

bin/cask
```diff
@@ -236,26 +236,48 @@ def find_best_emacs():
...
         return 'emacs'
 
+def inside_emacs_24():
+    """Whether Cask is running inside Emacs-24.
+
+Emacs 24 is rather inconsistent in its use of the ``$EMACS`` and
+``$INSIDE_EMACS`` environment variables, so we try here to test if Cask is
+being run inside Emacs (either in a shell or with ``M-x compile``).
+
+Return ``True`` when running in Emacs 24, ``False`` otherwise.
```
lunaryorn: Please fix the indentation of this docstring.

bin/cask
```diff
@@ -236,26 +236,48 @@ def find_best_emacs():
...
+    """
+    ## are we inside Emacs at all
+    return (b'INSIDE_EMACS' in ENVB and
+            ## M-x compile in Emacs-24 sets INSIDE_EMACS to exactly 't'
+            (ENVB.get(b'INSIDE_EMACS') == 't'
+             or
```
lunaryorn: Please join this line with the previous.

lunaryorn:
@rejeep @phillord There are two minor stylistic issues still. Please fix them and I'll merge.

bin/cask
swalec: Confused -- which line? docstring? Or the boolean?

phillord: Confused? Which line? The boolean or the last line of the docstring?

bin/cask
phillord: Everything to 4 spaces?

bin/cask
lunaryorn: @phillord Uhm, the line that I commented on? The one that only has `or` on it.

bin/cask
lunaryorn: Yes, like the other docstrings.  Emacs' built-in Python Mode automatically does that when you press `M-q` in a docstring.

bin/cask
phillord: Okay, no worries.

Does it say which line you commented on? I'm not getting a visual indication of that this end, just a snippet from line 236-252.

lunaryorn: Uhm, it does, doesn't it?

bildschirmfoto 2016-03-18 um 18 27 50

phillord: Not that I can see. Just the last line? Not that it matters -- I know which line you are talking about, and will fix.

phillord:
Should be done now.

bin/cask
lunaryorn: @phillord Uhm, I'm sorry but don't you see that my comment is associated with a line?!  Specifically the line right above my comment?!  I'm sorry but I thought that Github line comments were a known thing…

bin/cask
```diff
@@ -236,26 +236,48 @@ def find_best_emacs():
...
+    Return ``True`` when running in Emacs 24, ``False`` otherwise.
+
+    """
+    ## are we inside Emacs at all
+    return (b'INSIDE_EMACS' in ENVB and
+            ## M-x compile in Emacs-24 sets INSIDE_EMACS to exactly 't'
+            (ENVB.get(b'INSIDE_EMACS') == 't'
+             ## while M-x shell sets it to a string, starting with the
+             ## version number
+             or ENVB.get(b'INSIDE_EMACS').startswith('24')))
```
lunaryorn: @phillord I don't mean to annoy but I asked you to move the `or` up to the *previous* line, not onto the next one.  Just look how the line wraps *after* `and` in line 250.  When wrapping lines, boolean operators should go immediately *before* the line wrap.

lunaryorn:
@phillord Well, still not to my satisfaction, I'm sorry.  Please see my comment on the diff.  If this is addressed, I'll merge :)

phillord:
Fixed as requested.

lunaryorn:
@phillord Great, I merged this pull request :+1:  Thank you for your work, and for your patience with our questions, discussions and review comments :relaxed:

rejeep:
Awesome guys! :smile:

phillord:
Good stuff!

phillord:
Alas, on the day that this got merged, the EMACS=t behaviour was restored on Emacs-25. So, I fear, at least some of this is going to have to be backed out, and re-thought, unless I can get the restoration reverted.

Apologies for the hassle. I had no reason to expect that the Emacs 25 behaviour would be changed in pre-test.

lunaryorn:
@phillord Well, there you are.  Can we now finally settle on `$CASK_EMACS`, please?

phillord:
Working on it. The reasons for the restoration of EMACS=t are, I think, not that straight-forward. I have a solution either way, but I'll wait till the discussion has played out.

phillord:
All fixed. EMACS=t should be gone after all from Emacs-25.

rejeep:
@phillord so we don't need a revert?

phillord:
Indeed no. The intention is that Emacs will not set EMACS while will be reserved for "the Emacs I want to run", which is what Cask uses it for. There will be one exception to this (M-x term) which is unfortunate, which I'll try and fix for Emacs-25.2.

PR #337, opened 2016-03-03
Fix for Windows when cask path contains spaces.

bharath1097:
In windows if the cask install path contains spaces, cask doesn't work. This is because os.execvp can't handle spaces in path in windows.


rejeep:
@bharath1097 You closed this for a reason?

Issue #328, opened 2016-01-08
cask pkg-file shouldn't reorder dependencies


cpitclaudel:
`cask pkg-file` currently prints dependencies in reverse order (compared to the Package-Requires header). This can cause installation to fail on old emacsen (see https://github.com/milkypostman/melpa/issues/2744 for some context), so it would be great to respect the Package-Requires order :)

rejeep:
Pull request?

lunaryorn:
@cpitclaudel I guess there's an `nreverse` missing somewhere.  I for my part won't fix that any time soon though, because I'm not personally affected by it (I don't use older Emacsen), and there's more than enough work for me on the [Flycheck backlog](https://waffle.io/flycheck/flycheck) :blush:

cpitclaudel:
Happy to provide a patch. Looking at the code indicates two possible approaches:

* `reverse`ing in `cask--runtime-dependencies` and `cask-development-dependencies`
* Using `-snoc` instead of `push` in `cask-add-dependency`.

Does anyone have a preference for either option?

lunaryorn:
@cpitclaudel I would prefer the first one; it should be faster, because it only traverse `once`, whereas `-snoc` as to traverse for every single addition.

cpitclaudel:
Ok, I have a rough draft of this :)

PR #338, opened 2016-04-05
 Respect order of Package-Requires dependencies in cask pkg-file

cpitclaudel:
This is a rough draft: I'm still having trouble running tests; comments welcome!

Connects to #328

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -402,8 +402,7 @@ SCOPE may be nil or 'development."
          (cl-destructuring-bind (_ &rest patterns) form
            (setf (cask-bundle-patterns bundle) patterns)))
         (development
-         (cl-destructuring-bind (_ . body) form
-           (cask--eval bundle body 'development)))
+         (cask--eval bundle (cdr form) 'development))
         (t
          (error "Unknown directive: %S" form))))))
 
--- a/cask.el
+++ b/cask.el
@@ -564,8 +564,9 @@ This function return a `cask-bundle' object."
   (let* ((bundle
           (cask-setup (or project-path user-emacs-directory)))
          (package-load-list
-          (reverse (cons 'all (--map (list (cask-dependency-name it) t)
-                                     (cask--runtime-dependencies bundle))))))
+          (-snoc (--map (list (cask-dependency-name it) t)
+                        (cask--runtime-dependencies bundle))
+                 'all)))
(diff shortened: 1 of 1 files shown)


cpitclaudel:
I'd love help in understanding why that test is failing, too; it's not clear to me what part of the changes should affect it.

lunaryorn:
Well, changing the order of dependencies also changes the order in which immediate dependencies are installed, and thus also the order of packages in `package-alist` which tracks installed packages.  That's why the test is failing.

cpitclaudel:
Is that the case? I felt that that test was simpler than this; does it involve dependencies?

lunaryorn:
@cpitclaudel I think so.  Take a look at the test: It compiles a bundle with two direct dependencies (i.e. `depends-on`) and the _installs_ the bundle.

rejeep:
Instead of changing code in many places, why not update [the source](https://github.com/cask/cask/blob/master/cask.el#L451)?

cpitclaudel:
@rejeep I don't understand: that function isn't called if `deep` is `nil`.

rejeep:
Maybe I was mistaken, hard to read code on a mobile... 😄 Will check again later

cpitclaudel:
Thanks :)

cask.el
```diff
@@ -556,7 +561,8 @@ This function return a `cask-bundle' object."
   "Initialize packages under PROJECT-PATH or `user-emacs-directory'.
 
 This function return a `cask-bundle' object."
-  (let* ((bundle (cask-setup (or project-path user-emacs-directory)))
+  (let* ((bundle
+          (cask-setup (or project-path user-emacs-directory)))
```
rejeep: Did anything change here?

cask.el
```diff
@@ -402,8 +405,7 @@ SCOPE may be nil or 'development."
          (cl-destructuring-bind (_ &rest patterns) form
            (setf (cask-bundle-patterns bundle) patterns)))
         (development
-         (cl-destructuring-bind (_ . body) form
-           (cask--eval bundle body 'development)))
+         (cask--eval bundle (cdr form) 'development))
```
rejeep: Not related to this, right? In such case, please remove and make a separate pull request. We want to change as little as possible at a time.

rejeep:
Also, take a look at the tests, like @lunaryorn mentioned above.

Issue #339, opened 2016-04-08
Setting $EMACS environment variable does not work


paldepind:
Thank you for maintaining and developing Cask.

I am not sure if this is a bug in Cask. But when I set the `$EMACS` environment variable and run `cask exec ecukes` I get the following error:

```
$ cask exec ecukes
cask exec: error: [Errno 2] No such file or directory
```

I've tried both

```
EMACS="$(evm bin emacs-24.4-travis)"
```

And

```
EMACS="path/to/compiled/emacs"
```

I am using Cask version 0.7.4. Running Cask without the environment variable set works just fine.

rejeep:
Hey, we just merged some code related to the $EMACS env (see https://github.com/cask/cask/commit/9f101d91ad0a5a599dfbe3f64508acf70564c6f6). Does it work if you use $CASK_EMACS instead?

paldepind:
Setting `$CASK_EMACS` seems to have no effect.

```
$ CASK_EMACS="$(evm bin emacs-24.4-travis)" cask exec ecukes
24.5.1
23 scenarios (0 failed, 23 passed)
134 steps (0 failed, 0 skipped, 134 passed)
```

The version number being printed is being printed in the setup part of my tests and it shows the version of my system wide Emacs installation.

Should I try the current master?

phillord:
Are you running these command inside Emacs -- with M-x compile, or M-x shell? Or outside?

paldepind:
Outside of Emacs in a virtual terminal.

rejeep:
And you have installed the dependences for that version (they are scoped on version)? Take a look at `.cask`.

`EMACS="$(evm bin emacs-24.4-travis)" cask install`

paldepind:
Sorry. I did not do that. I though that the dependencies I had already installed would be used.

I've created #340 if you think this should be added to the documentation.

rejeep:
Thank you for those kind words! :smile: 

I figured it worked out for you? Can we close this?

paldepind:
Yes. Installing the dependencies worked. The only reason to leave it open would be that maybe the error message should be improved. `No such file or directory` is not that helpful.

rejeep:
I agree. I've seen others with this issue before. @lunaryorn would it be easy to put something like that in the bin-script?

lunaryorn:
@rejeep A better error message, that is?  I guess so, yes. I gave it a try, see d0dbaf5

rejeep:
> A better error message, that is?

[...]

PR #340, opened 2016-04-09
Note that dependencies are scoped on version

paldepind:
Mention that dependencies will have to be installed for each version of Emacs used.

See #339.

### DIFF
--- a/doc/guide/usage.rst
+++ b/doc/guide/usage.rst
@@ -52,6 +52,13 @@ the `emacs` command.  To pick a different Emacs, set the environment variable
 
    $ EMACS="emacs24.1" cask command
 
+Note that installed dependencies are scoped on the version of Emacs. So when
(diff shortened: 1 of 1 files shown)


doc/guide/usage.rst
```diff
@@ -52,6 +52,13 @@ the `emacs` command.  To pick a different Emacs, set the environment variable
 
    $ EMACS="emacs24.1" cask command
 
+Note that installed dependencies are scoped on the version of Emacs. So when
+switching between versions you will have to install the dependencies for each:
+
+.. code-block:: console
+
+   $ EMACS="emacs24.1" cask install
```
rejeep: Perhaps we should use a different version from the one above here?

paldepind:
> Perhaps we should use a different version from the one above here?

Great idea. I've changed it to a different version.

rejeep:
Thanks! 👍

PR #347, opened 2016-04-30
Provide emacs-with-cask

phillord:
Allow running an emacs setup to use local cask settings
but with full access to other command line arguments.

Not a real PR at the moment, just thinking out load. I'm wondering whether the emacs command I have added is quite the right thing. It's useful, but it doesn't set up the environment as an Emacs running cask would.

This version (emacs-with-cask -- bad name) does this. It allows me to do things like

    cask emacs-with-cask -q --batch --eval '(message "hello world")'

which replicates the existing `cask eval` command. But I can also do

    cask emacs-with-cask -q --batch --load blah.el -f blah-run-everything-interesting

So, instead of adding as well as "eval", a load-file, a run function and other options to cask, we just let the Emacs command line parser do all the work. We can't get this functionality at the moment AFAICT, with the `emacs` command because it doesn't know where `cask.el` is.

Perhaps there is a better way to do this, and I am missing something that cask already does. Other thoughts -- is "emacs-with-cask" sensible, and perhaps this should just be the behaviour of the "emacs" command. And, if emacs-with-cask is sensible, this is probably the right time to put a proper python command line parser into cask. Or, perhaps, just add CASK_DIRECTORY to the environment of the Emacs command and let people do this for themselves.

Obviously, docstrings, manual and (this time) tests would be added also.

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -337,6 +337,13 @@ def exec_emacs(command):
     exec_command([get_cask_emacs()] + command)
 
 
+def exec_emacs_with_cask(command):
+    ENVB[b'CASK_DIRECTORY'] = CASK_DIRECTORY
+    exec_command([get_cask_emacs(), "--load"] +
+                 ["{}/cask-init.el".format(CASK_DIRECTORY)] +
+                 command)
+
+
 def exec_cask(arguments):
     """Execute the Cask CLI with the given ``arguments``.
 
@@ -388,6 +395,8 @@ def main():
             exec_command(sys.argv[2:])
         elif len(sys.argv) > 1 and sys.argv[1] == 'emacs':
             exec_emacs(sys.argv[2:])
+        elif len(sys.argv) > 1 and sys.argv[1] == "emacs-with-cask":
+            exec_emacs_with_cask(sys.argv[2:])
         else:
             exec_cask(sys.argv[1:])
     except OSError as error:
--- a/cask-init.el
+++ b/cask-init.el
@@ -1,0 +1,41 @@
+;;; cask-init.el --- Set up an Emacs
+
+;; Copyright (C) 2016 Phillip Lord
+
+;; Author: Phillip Lord <phillip.lord@russet.org.uk>
+;; Maintainer: Phillip Lord <phillip.lord@russet.org.uk>
+;; URL: http://github.com/cask/cask
+
+;; This file is NOT part of GNU Emacs.
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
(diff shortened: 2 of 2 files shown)


rejeep:
I think that eval, emacs and emacs-with-cask should be three different things. If we add a proper cli parser, we could use an API like this: `cask emacs --initialize ...`

I wouldn't use this nor the emacs command as is very much. What is your real use case for these commands, other than printing _"hello world"_. 😄

phillord:
Actually, I am not sure that I do need it. My impression was that `cask emacs` was not including packages in `.cask` in the load-path. But this was happening because I had updated emacs version and not run `cask install`. Perhaps `cask` should give an error message in this case, rather than carrying on.

My use case is that I want to do something like

      cask emacs --batch -q -l assess-discover.el -f assess-batch-run-tests-and-exit

which finds all local tests and runs them. I also have things like:

    $(CASK) emacs -q --batch --script script/build.el -- gen-html

In this case, build.el generates some HTML from source code using my own lentic package and then org. build.el uses commander to call the relevant function, which just seens unnecessary when I could just do:

    cask emacs -q --batch --load script/build.el -f generate-html

The caveat to this is that `cask emacs` does not include the default-directory in load-path, which this PR also adds.

So short summary: 1) Should `cask emacs` (and `eval` and `exec`) signal errors if `cask install` has not been run. 2) should `cask emacs` (and `eval`) include the CWD in EMACSLOADPATH.

lunaryorn:
Sure, 1) would be a good idea, but I don't that 2) is.  Emacs itself doesn't add the working directory to `load-path` either, and we shouldn't add non-standard on top of Emacs, all the more since it'd be somewhat unusual behaviour.  Shells don't add the working directory `$PATH` either.

phillord:
@lunaryorn I'll add 1) in a different PR. 

On 2), cask defines a project environment which includes lots of files in the current working directory. Most package tools have default locations for finding source files relative to their project file (maven has `src/java`, lein uses `src` and so forth).

Currently, if I load a package with -l from the local project, then inter-project `require` statements will fail, which seems unfortunate.

In the idea world, I guess, adding the files in the Cask file would be the way forward. But I know of no way of doing that short of loading them.

rejeep:
1) 👍 

2) Not sure about this one. Packages such as ert-runner, ecukes, ... can always decide to do this on their own, or add an option for it, like [ert-runner](https://github.com/rejeep/ert-runner.el/blob/master/ert-runner.el#L338).

phillord:
1) Other PR sent.
2) Okay, no worries. It can be done on the Emacs command line anyway...

	$(EMACS_ENV) $(CASK) emacs --batch -q \
	--directory=. \
	--load assess-discover.el \
	--funcall assess-discover-run-and-exit-batch


I'll close this one.

rejeep:
Maybe we'll do something about (2) if enough people complain.

PR #349, opened 2016-05-02
Add error if cask install has not been run

phillord:
Cask now checks for the presence of an ELPA directory, and signals an
error if it does not exist for all commands where the ELPA directory
will change the result of the output.

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -90,6 +90,19 @@ class MissingEmacsError(Exception):
         self.emacs = emacs
 
 
+class MissingPackageError(Exception):
+    """An exception indicating that one or more direct dependencies are missing.
+
+    The ```report``` attribute describes the missing dependencies in a
+    human-readable form.
+
+    """
+
+    def __init__(self, report):
+        Exception.__init__(self)
+        self.report = report
+
+
 class EmacsCommandError(Exception):
     """An exception indicating a failed Emacs command.
 
@@ -153,18 +166,35 @@ def format_version(version):
     return '.'.join(str(part) for part in version)
 
 
+def get_missing_packages():
+    """Get a report of the missing direct dependencies.
+
+    The report is a human-readable list of the any direct dependencies which are
+    required but not present. The empty string is returned if there are none.
+
+    """
+    process = subprocess.Popen([sys.executable, CASK, "missing"], stdout=subprocess.PIPE)
+    stdout, _ = process.communicate()
+
+    return stdout.rstrip()
+
+
 def get_cask_path(kind):
     """Get the Cask package environment path of the given ``kind``.
 
     ``kind`` is a string, denoting the kind of the path to get.  Cask supports
-    two different kinds, namely ``'path'`` for the executable path of the
-    package environment, and ``'load-path'`` for the Emacs Lisp load path of
-    the package environment.
+    three different kinds, namely ``'path'`` for the executable path of the
+    package environment, ``'load-path'`` for the Emacs Lisp load path of
+    the package environment and ``'package-directory'``.
 
     Return the path as string, which contains all directories of the path
     separated by path separators.
 
     """
+    missing = get_missing_packages()
+    if (missing):
+        raise MissingPackageError(missing)
+
     process = subprocess.Popen([sys.executable, CASK, kind], stdout=subprocess.PIPE)
     stdout, _ = process.communicate()
     return stdout.rstrip()
@@ -400,6 +430,8 @@ def main():
         exit_error(
             'Emacs does not exist at {0}.  Did you install Emacs?'.format(
                 error.emacs))
+    except MissingPackageError as error:
+        exit_error("Missing Packages\n{}\nDo you need to run cask install?".format(error.report))
     except EmacsCommandError as error:
         exit_error('Emacs command "{0}" failed: {1}'.format(
             error.command, error.real_error))
--- a/cask-cli.el
+++ b/cask-cli.el
@@ -87,6 +87,33 @@
           (princ (s-pad-right (+ max cask-cli--table-padding) " " key))
           (princ (concat value "\n")))))))
 
+(defun cask-cli--missing-dependencies ()
+  "Return a list of direct dependencies that are missing."
+  (let ((bundle (cask-cli--bundle)))
+    (-remove
+     (lambda (dependency)
+       (let ((name (cask-dependency-name dependency)))
+         (or (epl-package-installed-p name)
+             (cask-linked-p bundle name))))
+     (cask-dependencies bundle))))
+
+(defun cask-cli--missing-dependencies-report (missing)
+  "Return a string report describing missing dependencies."
+  (->> missing
+       (-map #'cask-dependency-name)
+       (-map #'symbol-name)
+       (s-join "\n")))
+
+(defun cask-cli--dependencies-exist-or-error ()
+  "Check if all direct dependencies exist.
+
+Signal an error if not."
+  (-when-let (missing-dependencies (cask-cli--missing-dependencies))
+    ;; This error message should be the same as the one from bin/cask.
+    ;; Any changes need to be percolated.
+    (error "Missing Dependencies.\n%s\nRun cask install to update."
+           (cask-cli--missing-dependencies-report missing-dependencies))))
+
 
 ;;;; Commands
 
@@ -164,6 +191,7 @@ Git is available in `exec-path'."
 
 All packages that are specified in the Cask-file will be updated
 including their dependencies."
+  (cask-cli--dependencies-exist-or-error)
   (cask-cli/with-handled-errors
     (-when-let (upgrades (cask-update (cask-cli--bundle)))
       (princ "Updated packages:\n")
@@ -213,6 +241,7 @@ will be for an Emacs package."
   "Print `load-path' for all packages and dependencies.
 
 The output is formatted as a colon path."
+  (cask-cli--dependencies-exist-or-error)
   (princ (concat (s-join path-separator (cask-load-path (cask-cli--bundle))) "\n")))
 
 (defun cask-cli/exec-path ()
@@ -222,6 +251,7 @@ A dependency will be included in this list of the package has a
 directory called bin in the root directory.
 
 The output is formatted as a colon path."
+  (cask-cli--dependencies-exist-or-error)
   (princ (concat (s-join path-separator (cask-exec-path (cask-cli--bundle))) "\n")))
 
 (defmacro cask-cli--with-package-path (&rest body)
@@ -235,6 +265,7 @@ The output is formatted as a colon path."
 
 (defun cask-cli/eval (form)
   "Eval FORM with the `load-path' set according to the project."
+  (cask-cli--dependencies-exist-or-error)
   (cask-cli--with-package-path
    (eval (read form))))
 
@@ -247,9 +278,16 @@ The output is formatted as a colon path."
 
 That is packages that have a more recent version available for
 installation."
+  (cask-cli--dependencies-exist-or-error)
   (-when-let (outdated (cask-outdated (cask-cli--bundle)))
     (princ "Outdated packages:\n")
     (-each outdated 'cask-cli--print-upgrade)))
+
+(defun cask-cli/missing ()
+  "Print a list of direct dependencies which are missing."
+  (-when-let (missing (cask-cli--missing-dependencies))
+    (princ (cask-cli--missing-dependencies-report missing))
+    (princ "\n")))
 
 (defun cask-cli/files ()
   "Print list of files specified in the files directive.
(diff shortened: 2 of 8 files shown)


lunaryorn:
Thank you very much.  I'm afraid, though, that this is rather not what I'd hoped for.  Checking the directory isn't really good enough, and will only catch a very narrow set of errors—essentially just when you switch to a new Emacs version.

But what if you add `ert-runner` to your `Cask` file and run `cask exec ert-runner`?  You'll still get the very same error, because the directory exists, but not the particular dependency you're interested in.

If we verify whether dependencies are installed, we should really, well, verify that dependencies are actually installed, i.e. that all the dependencies specified in the `Cask` file are actually installed.

phillord:
Yeah, switching Emacs is something I do a lot which is probably why this bothers me. No worries I will fix. I'll have to add a new command to cask, though, I think. I can duplicate checking whether a directory exists in python, but not whether all dependencies exist.

lunaryorn:
@phillord That's fine, as far as I'm concerned.  We should probably have a `cask verify` command anyway to check whether all deps are available.

rejeep:
It still wouldn't be perfect because ert-runner can have dependencies, which may not be installed. But I guess it's good enough.

It's not used very often, but the second argument to `depends-on` is a version. We should use that when checking that a package has been installed.

> We should probably have a cask verify command anyway to check whether all deps are available.

Bundle has a command for that http://bundler.io/v1.1/bundle_check.html

phillord:
I have updated. It only checks direct dependencies AFAICT, but this seems fine to me, as transitive dependencies would require all the packages to be present.

Of course, there is a simpler option to all of this -- just run "cask install" instead of erroring. That would also be a simply way of doing "verify".

bin/cask
```diff
@@ -153,20 +166,36 @@ def format_version(version):
...
 
 
+def get_missing_packages():
+    """Get a report of the missing direct dependencies.
+
+    The report is a human-readable list of the any direct dependencies which are
+    required but not present. The empty string is returned if there are none.
+
+    """
+    process = subprocess.Popen([sys.executable, CASK, "missing"], stdout=subprocess.PIPE)
```
lunaryorn: Please check the return code of the process and raise an exception if it's not zero.  We should fail when we fail to get the list of packages, instead of silently printing something.

Furthermore, I do not like that this method returns a string, let alone a "human-readable" string.  It should return a list of missing packages, or—probably more appropriate in our case—a boolean indicating whether packages are missing or not (though in this case the function should be called `are_packages_missing` or so).

bin/cask
```diff
@@ -153,20 +166,36 @@ def format_version(version):
...
     separated by path separators.
 
     """
+    missing = get_missing_packages()
+    if(missing):
+        raise MissingPackageError(missing)
+
     process = subprocess.Popen([sys.executable, CASK, kind], stdout=subprocess.PIPE)
-    stdout, _ = process.communicate()
+    stdout, stderr = process.communicate()
```
lunaryorn: What do we need `stderr` for here?

bin/cask
```diff
@@ -295,7 +324,6 @@ def get_cask_emacs():
     ensure_supported_emacs(emacs)
     return emacs
 
-
```
lunaryorn: Please restore this blank line.

cask-cli.el
```diff
@@ -87,6 +87,33 @@
...
+         (or (epl-package-installed-p name)
+             (cask-linked-p bundle name))))
+     (cask-dependencies bundle))))
+
+(defun cask-cli--missing-dependencies-report (missing)
+  "Return a string report describing missing dependencies."
+  (->> missing
+       (-map #'cask-dependency-name)
+       (-map #'symbol-name)
+       (s-join ", ")))
```
lunaryorn: Please use a line break to separate the list of missing packages.  That's easier to read I think, and it's easier to feed into other programs like `grep` or whatever.

bin/cask
phillord: Return code, yes, will add.

Returning a structured string or a boolean, I thought about that. The problem is that we can get the same error from inside python (cask exec, cask emacs) or inside Emacs (cask outdated, eval and several others). Seems to me that the error message should look the same in both cases. Otherwise, I have to return a parsable list from cask missing, then print it to look the same as the error message that we would get from lisp.

cask-cli.el
rejeep: Use https://github.com/cask/cask/blob/master/cask-cli.el#L62

cask-cli.el
```diff
@@ -87,6 +87,33 @@
...
+
+(defun cask-cli--elpa-dependencies-exist-or-error ()
+  "Check if local ELPA directory exists.
```
rejeep: This description is outdated.

cask-cli.el
```diff
@@ -87,6 +87,33 @@
           (princ (s-pad-right (+ max cask-cli--table-padding) " " key))
           (princ (concat value "\n")))))))
 
+(defun cask-cli--missing-elpa-dependencies ()
```
rejeep: I don't think we should use _"elpa"_ internally. Everything in Cask is based on elpa and there are no other dependencies than elpa dependencies.

cask-cli.el
```diff
@@ -247,10 +278,17 @@ The output is formatted as a colon path."
...
 installation."
+  (cask-cli--elpa-dependencies-exist-or-error)
   (-when-let (outdated (cask-outdated (cask-cli--bundle)))
     (princ "Outdated packages:\n")
     (-each outdated 'cask-cli--print-upgrade)))
 
+(defun cask-cli/missing ()
+  "Print a list of direct dependencies which are missing."
+  (-when-let (missing (cask-cli--missing-elpa-dependencies))
+    (princ "Missing packages:\n")
```
rejeep: I think we should remove this header. Like @lunaryorn said, it's easier to grep on a "clean" list.

bin/cask
```diff
@@ -153,18 +166,38 @@ def format_version(version):
...
+
+    The report is a human-readable list of the any direct dependencies which are
+    required but not present. The empty string is returned if there are none.
+
+    """
+    process = subprocess.Popen([sys.executable, CASK, "missing"], stdout=subprocess.PIPE)
+    stdout, _ = process.communicate()
+
+    if(process.returncode):
+        raise Exception("Oops")
```
phillord: I am a little unclear what you want me to do here. Just throw an error? Mostly this code is copied from `get_cask_path` which has no handling for the broken case.

lunaryorn: Oh, well, that's too bad for `get_cask_path` ;).  Leave it out then and just do it like `get_cask_path`.  I thought that we'd have checked these things for Cask functions but I've been mistaken.

Sorry.

phillord:
Outstanding issues.

I've removed the header from `missing`. While there is sense in this when "missing" is being used internally, it makes it less good as a command (compare with `outdated`). Possibly that is not a problem -- who is going to use `missing` interactively, when `cask install` just solves the problem.

`get_missing_packages` still returns (a slight less) human-readable report. @lunaryorn  are you happy with my reasoning here, or do you want to change it?

rejeep:
> it makes it less good as a command (compare with outdated)

Yeah, so perhaps we should revice outdated instead. Would be good if the output format was the same.

phillord:
We can square the circle by adding a `--plumbing` option. Then there would not be a conflict between clear messages for users, parsable messages for tools (including `cask`)

phillord:
We happy with this now?

bin/cask
```diff
@@ -153,18 +166,35 @@ def format_version(version):
...
+    three different kinds, namely ``'path'`` for the executable path of the
+    package environment, ``'load-path'`` for the Emacs Lisp load path of
+    the package environment and ``'package-directory'``.
 
     Return the path as string, which contains all directories of the path
     separated by path separators.
 
     """
+    missing = get_missing_packages()
+    if(missing):
```
lunaryorn: A space before `(`, please.

bin/cask
```diff
@@ -90,6 +90,19 @@ class MissingEmacsError(Exception):
...
 
+class MissingPackageError(Exception):
+    """An exception indicating that one or more direct dependencies are missing.
+
+    The ```report``` attribute describes the missing dependencies in a
+    human-readable form.
+
+    """
+
+    def __init__(self,report):
```
lunaryorn: A space after `, ` please :)

lunaryorn:
@phillord The Python side looks good to me, except of some minor style issues which I'd like to ask you to fix.

I've no particular opinion about the CLI interface, and let @rejeep make the decision here.

phillord:
@lunaryorn All done.

Been meaning to ask, do you use pylint or some such? I should stop hassling you with these code corrections, but I don't have my eye in with python layout as well as you seem to?

@rejeep If you are happy, please feel to merge. I can add a `--plumbing` option in a separate commit, and do any other commands that might need it.

lunaryorn:
@phillord I do use pylint for all of my code, but having written Python code for almost a decade I can't really see over these style glitches anymore :blush:

phillord:
@lunaryorn I'll add pylint to my workflow! I use too many languages to have all of the idioms burnt into my brain.

rejeep:
@phillord What would the plumbing option do and what commands would it be available for?

Also, please add some docs for the new `missing` command.

phillord:
@rejeep Doc added.

Haven't thought about `--plumbing` in detail, and think it is for a separate day. But for user commands, I don't like worrying about "easy-to-parse", only "easy-to-read".

features/emacs.feature
```diff
@@ -6,7 +6,8 @@ Feature: Eval
       """
       (package "super-project" "0.0.1" "Super project.")
       """
-    When I run cask "emacs -q --batch --eval '(print nil)'"
+    When I run cask "install"
+    and I run cask "emacs -q --batch --eval '(print nil)'"
```
rejeep: Capitalize

features/missing.feature
```diff
@@ -0,0 +1,43 @@
...
+    (depends-on "package-b" "0.2.1")
+      
+    (development
+      (depends-on "package-c")
+      (depends-on "package-d"))
+    """
+    When I run cask "list"
+    Then I should see command output:
+    """
+    """
```
rejeep: I guess you don't want the empty string here!? 😄 

Also, can you add a test for a linked package.

features/missing.feature
```diff
@@ -0,0 +1,43 @@
...
+    
+    (development
+    (depends-on "package-c")
+    (depends-on "package-d"))
+    """
+    When I run cask "install"
+    And I run cask "missing"
+    Then I should see command output:
+    """
+    """
```
rejeep: This will always be true. It would be better with "Then I should not see command output"

features/missing.feature
```diff
@@ -0,0 +1,43 @@
...
+    Given this Cask file:
+    """
+    (depends-on "package-a" "0.1.2")
+    (depends-on "package-b" "0.2.1")
+      
+    (development
+      (depends-on "package-c")
+      (depends-on "package-d"))
+    """
+    When I run cask "list"
```
rejeep: "list" => "missing"

features/missing.feature
```diff
@@ -0,0 +1,43 @@
...
+    
+  Scenario: With dependencies
+    Given this Cask file:
+    """
+    (depends-on "package-a" "0.1.2")
+    (depends-on "package-b" "0.2.1")
+    
+    (development
+    (depends-on "package-c")
+    (depends-on "package-d"))
```
rejeep: Indentation

features/missing.feature
```diff
@@ -0,0 +1,43 @@
...
+  List missing dependencies
+
+  Scenario: No dependencies
+    Given this Cask file:
+    """
+    """
+    When I run cask "missing"
+    Then I should see command output:
+    """
+    """
```
rejeep: Same here as above. It will always be true.

When writing specs. Make sure they fail without the changes in the code. Otherwise you can never trust that the test is correctly implemented.

phillord:
Mostly done -- not sure about the test for a linked package -- what are you after?

rejeep:
> what are you after?

What I was thinking about was, what happens if you run `cask install`, then link a package and then run `cask missing`. Will that package be considered installed or not? A test for that case would be good.

Other than that I'm happy.

phillord:
Hmmm, had a go at checking on the behaviour of `cask link`. On the command line it seems to work as you might want, but I can't get my test working -- or rather it fails when I think it should pass. Got any ideas?

phillord:
Am rather struggling with this one -- can't think how to test `cask link` better -- I need something to link to.

Suggest I remove this test. I can revisit this test if when more fixtures have been added for `cask linked`.

rejeep:
I'm fine with skipping the link test, but please make sure it works.

PR #351, opened 2016-05-08
Exit build command with proper status code

rejeep:
If any of the files contained a compiler error, exit with status code 1, otherwise 0.

Also add the option `--warn-as-error` that will consider compiler warnings as errors, which means if any of the files contained any warning, exit with status code 1.

### DIFF
--- a/cask-cli.el
+++ b/cask-cli.el
@@ -52,6 +52,9 @@
 (defvar cask-cli--path default-directory
   "Cask commands will execute in this path.")
 
+(defvar cask-cli--warn-as-error nil
+  "Consider warnings as error.")
+
 (defvar cask-cli--bundle-cache nil)
 
 (defun cask-cli--bundle ()
@@ -260,8 +263,14 @@ If no files directive or no files, do nothing."
       (princ (concat file "\n")))))
 
 (defun cask-cli/build ()
-  "Build all Elisp files in the files directive."
-  (cask-build (cask-cli--bundle)))
+  "Build all Elisp files in the files directive.
+
+Exit code is 1 if any of the files contained a compile error, 0
+otherwise."
+  (let ((result (cask-build (cask-cli--bundle) :warn-as-error cask-cli--warn-as-error)))
+    (if result
+        (kill-emacs 0)
+      (kill-emacs 1))))
 
 (defun cask-cli/clean-elc ()
   "Remove all byte compiled Elisp files in the files directive."
@@ -357,6 +366,10 @@ Commands:
   "Be verbose and do not hide output."
   (setq shut-up-ignore t))
 
+(defun cask-cli/warn-as-error ()
+  "Consider compiler warnings as errors."
+  (setq cask-cli--warn-as-error t))
+
 
 ;;;; Commander schedule
 
@@ -399,7 +412,8 @@ Commands:
  (option "--dev" cask-cli/dev)
  (option "--debug" cask-cli/debug)
  (option "--path <path>" cask-cli/set-path)
- (option "--verbose" cask-cli/verbose))
+ (option "--verbose" cask-cli/verbose)
+ (option "--warn-as-error" cask-cli/warn-as-error))
 
 (provide 'cask-cli)
 
--- a/cask.el
+++ b/cask.el
@@ -833,17 +833,29 @@ URL is the url to the mirror."
                            (cask-bundle-sources bundle))))
     (setf (cask-bundle-sources bundle) sources)))
 
-(defun cask-build (bundle)
-  "Build BUNDLE Elisp files."
+(defun cask-build (bundle &rest args)
+  "Build BUNDLE Elisp files.
+
+Return t if all files were successfully built without errors,
+otherwise return nil.
+
+ARGS supports the following options:
+ - warn-as-error: If t, then warnings are considered errors."
   (cask--with-file bundle
     (require 'bytecomp)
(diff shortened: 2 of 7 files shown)


yuutayamada:
Thank you for your quick implement.

I could make sure warnings become error and returning status code 1, but
when there is only an error, cask still returns status code 0.

I tried following file foo.el
(I tried both `cask build` and `cask build --warn-as-error`. and
result was same)

```el
;; -*- lexical-binding: t; -*-

;;; Code:

(defun foo ()
  (let* ((a nil))
    (add-to-list 'a "foo")))

(provide 'foo)

;; Local Variables:
;; coding: utf-8
;; mode: emacs-lisp
;; End:

;;; foo.el ends here
```

it output:
```sh
§~/local/vcs/github.com/yuutayamada/emacs.d/elisp/self/hoge§
λ cask build
Compiling /home/uta/local/vcs/github.com/yuutayamada/emacs.d/elisp/self/hoge/foo.el...

In foo:
foo.el:6:13:Error: `add-to-list' can't use lexical var `a'; use `push' or
    `cl-pushnew'
```

and return code was:
```sh
§uta@Penguin§☁|origin/master§2016/05/08 14:01§
§~/local/vcs/github.com/yuutayamada/emacs.d/elisp/self/hoge§
λ echo $?
0
```

rejeep:
That's weird. For some reason `byte-recompile-file` return `t` for `foo.el`.

yuutayamada:
I also got `t` from `byte-recompile-file` and when I tried another code that
occur an error, it return nil. (and success to return status code 1 in `cask build`)

I tried this code:
```lisp
(require ')
```

maybe due to emacs' bug? I'll ask emacs developers. (I could reproduce this emacs'
master branch as well)

Anyway thank you for implementing this feature!

rejeep:
Great, let us know what the devs says about it.

cask-cli.el
```diff
@@ -260,8 +263,14 @@ If no files directive or no files, do nothing."
...
-  "Build all Elisp files in the files directive."
-  (cask-build (cask-cli--bundle)))
+  "Build all Elisp files in the files directive.
+
+Exit code is 1 if any of the files contained a compile error, 0
+otherwise."
+  (let ((result (cask-build (cask-cli--bundle) :warn-as-error cask-cli--warn-as-error)))
+    (if result
+        (kill-emacs 0)
+      (kill-emacs 1))))
```
lunaryorn: What's `result` here?  Which type is it?

And please do `(kill-emacs (if result 0 1))`…

cask.el
```diff
@@ -833,17 +833,29 @@ URL is the url to the mirror."
                            (cask-bundle-sources bundle))))
     (setf (cask-bundle-sources bundle) sources)))
 
-(defun cask-build (bundle)
-  "Build BUNDLE Elisp files."
+(defun cask-build (bundle &rest args)
+  "Build BUNDLE Elisp files.
+
+Return t if all files were successfully built without errors,
+otherwise return nil.
```
lunaryorn: If we want Cask to exit with failure on compile errors we should use `error` to really abort compilation and not just return nil/non-nil.  IOW, we should not propagate exceptional conditions with return values.

Please refactor the code to use `error`s for compile errors.  If need be define a custom error signal (which I think we already do for Cask's own errors).

cask.el
```diff
@@ -833,17 +833,29 @@ URL is the url to the mirror."
...
+
+ARGS supports the following options:
+ - warn-as-error: If t, then warnings are considered errors."
```
lunaryorn: That would be non-nil and not just `t`.

cask.el
```diff
@@ -833,17 +833,29 @@ URL is the url to the mirror."
                            (cask-bundle-sources bundle))))
     (setf (cask-bundle-sources bundle) sources)))
 
-(defun cask-build (bundle)
-  "Build BUNDLE Elisp files."
+(defun cask-build (bundle &rest args)
```
lunaryorn: The argument should be `properties` or `plist`, because that's what it is :)

lunaryorn:
@rejeep I don't like using nil/non-nil return values for `cask-build`.  If we want compile errors to trigger failure, we should, well, fail.  Signal an error, that is.

yuutayamada:
just FYI, I conformed the bug? in [here](http://debbugs.gnu.org/cgi/bugreport.cgi?bug=23495).

yuutayamada:
@rejeep @lunaryorn I conformed the bug (emit warning, but doesn't return error status code when we set --warn-as-error) to emacs devel, but I seem it will take more time until I get answer or the fix. On the other hand, this PR give us more reliability for testing, so could you guys proceed this PR to be merged?

rejeep:
@yuutayamada Sorry, have been on paternity leave for a while and have not done any programming. Will be back to normal soon, will look into it then.

vermiculus:
Just a friendly ping; this functionality would be incredibly useful for integration testing 😄

rejeep:
@vermiculus If this is something you want in master, can you please rebase to master and verify that it still works (was a really long time ago). If all is well I can merge it!

vermiculus:
@rejeep I've added it to my TODOs for this weekend 👍

rejeep:
Great! Ping me when you are done and I'll help you out!

vermiculus:
I am working on this, but how do you want me to submit the patch? Should another PR be opened and this one closed? Seems…dirty.

rejeep:
@vermiculus Do it as you prefer!

PR #355, opened 2016-06-09
Fixes #252 add support for package pinning

### LINKED ISSUE #252: Add support for pinned packages
package.el in Emacs 24.4 includes a new feature to pin packages (see `package-pinned-packages`). Provide a way for Cask to pin packages.

For example:

``` lisp
(depends-on "foo")
(depends-on "bar" :archive "melpa-stable")
```

dgellow:
Fixes #252

I added an `archive` attribute to the `cask-dependency` cl-struct.
I read it from a `:archive "melpa-stable"` specified in the `(depends-on )` form, set the field in the dependency cl-struct and update `package-pinned-packages`.

Example:

```
(depends-on "f" :archive "melpa-stable")
```

Note: `package-pinned-packages` is from emacs v24.4 if C-h v is correct.

~~Do not merge, I still need to update the documentation and write some tests.~~
Btw I don't write emacs lisp often, if someone could proofread that would be great.

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -815,7 +815,7 @@ url to the fetcher source."
     (-when-let (archive (plist-get args :archive))
       (if (not (boundp 'package-pinned-packages))
           (princ "Archive pinning is only supported with Emacs v24.4+.\n")
-          (do
+          (progn
               (setq package-pinned-packages
                     (add-to-list
                      'package-pinned-packages
--- a/cask.el
+++ b/cask.el
@@ -106,8 +106,10 @@ when fetcher is specified.
 
 `ref' The ref to use if fetcher.
 
-`branch' The branch to use if fetcher."
-  name version fetcher url files ref branch)
+`branch' The branch to use if fetcher.
+
+`archive' The archive to pin the dependency to."
+  name version fetcher url files ref branch archive)
 
 (cl-defstruct cask-source
   "Structure representing a package source.
@@ -792,6 +794,8 @@ ARGS is a plist with these optional arguments:
 
  `:branch' Fetcher branch to checkout.
 
+ `:archive' Pin package to archive.
+
 ARGS can also include any of the items in `cask-fetchers'.  The
 plist key is one of the items in the list and the value is the
 url to the fetcher source."
@@ -808,6 +812,12 @@ url to the fetcher source."
         (setf (cask-dependency-ref dependency) ref))
       (-when-let (branch (plist-get args :branch))
         (setf (cask-dependency-branch dependency) branch)))
+    (-when-let (archive (plist-get args :archive))
+      (setq package-pinned-packages
+            (add-to-list
+             'package-pinned-packages
+             `(,name . ,archive) package-pinned-packages))
+      (setf (cask-dependency-archive dependency) archive))
     (if (eq (plist-get args :scope) 'development)
         (push dependency (cask-bundle-development-dependencies bundle))
       (push dependency (cask-bundle-runtime-dependencies bundle)))))
--- a/cask.el
+++ b/cask.el
@@ -86,6 +86,7 @@ Defaults to `error'."
 (define-error 'cask-failed-initialization "Failed initialization" 'cask-error)
 (define-error 'cask-not-a-package "Missing `package` or `package-file` directive" 'cask-error)
 (define-error 'cask-no-cask-file "Cask file does not exist" 'cask-error)
+(define-error 'cask-no-pinning "Package pinning is only supported with Emacs v24.4+." 'cask-error)
 
 (cl-defstruct cask-dependency
   "Structure representing a dependency.
@@ -813,14 +814,13 @@ url to the fetcher source."
       (-when-let (branch (plist-get args :branch))
         (setf (cask-dependency-branch dependency) branch)))
     (-when-let (archive (plist-get args :archive))
-      (if (not (boundp 'package-pinned-packages))
-          (princ "Archive pinning is only supported with Emacs v24.4+.\n")
-          (progn
-              (setq package-pinned-packages
-                    (add-to-list
-                     'package-pinned-packages
-                     `(,name . ,archive) package-pinned-packages))
-              (setf (cask-dependency-archive dependency) archive))))
+      (unless (boundp 'package-pinned-packages)
+        (signal 'cask-no-pinning nil))
+      (setq package-pinned-packages
+            (add-to-list
+             'package-pinned-packages
+             `(,name . ,archive) package-pinned-packages))
+      (setf (cask-dependency-archive dependency) archive))
     (if (eq (plist-get args :scope) 'development)
         (push dependency (cask-bundle-development-dependencies bundle))
       (push dependency (cask-bundle-runtime-dependencies bundle)))))
--- a/cask.el
+++ b/cask.el
@@ -794,7 +794,7 @@ ARGS is a plist with these optional arguments:
 
  `:branch' Fetcher branch to checkout.
 
- `:archive' Pin package to archive.
+ `:archive' Pin package to archive. Only supported by Emacs v24.4+.
 
 ARGS can also include any of the items in `cask-fetchers'.  The
 plist key is one of the items in the list and the value is the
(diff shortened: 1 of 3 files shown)


dgellow:
Should be ok now.

cask.el
```diff
@@ -806,6 +810,15 @@ url to the fetcher source."
         (setf (cask-dependency-ref dependency) ref))
       (-when-let (branch (plist-get args :branch))
         (setf (cask-dependency-branch dependency) branch)))
+    (-when-let (archive (plist-get args :archive))
+      (if (not (boundp 'package-pinned-packages))
+          (princ "Archive pinning is only supported with Emacs v24.4+.\n")
```
rejeep: Raise a proper error here. We do this all over the place in the code so take a look at how we do there.

test/cask-api-test.el
```diff
@@ -630,6 +630,19 @@
       (cask-install bundle)
       (should (equal orig-load-path load-path)))))
 
+(ert-deftest cask-install-test/pin-archive ()
+  (if  (not (boundp 'package-pinned-packages))
+    (princ "`package-pinned-packages' is not defined, we need Emacs v24.4+ to test the pinning feature.\n")
+    (cask-test/with-bundle
+     '((source localhost)
+       (depends-on "package-a" "0.0.1" :archive "localhost"))
```
rejeep: I'm not sure if we should use a string here if the source is one of the [supported ones](https://github.com/cask/cask/blob/f9e804ea96d14e24ae49eeb72d95d44e358b38f2/cask.el#L159-L166). @lunaryorn What do you think?

test/cask-api-test.el
```diff
@@ -630,6 +630,19 @@
       (cask-install bundle)
       (should (equal orig-load-path load-path)))))
 
+(ert-deftest cask-install-test/pin-archive ()
+  (if  (not (boundp 'package-pinned-packages))
+    (princ "`package-pinned-packages' is not defined, we need Emacs v24.4+ to test the pinning feature.\n")
```
rejeep: See my other comment.

rejeep:
Change the current test and "mock" the `package-pinned-packages` and make sure it works. Then write another test where you unset the variable and make sure it raises (`should-error`).

cask.el
dgellow: My approach was to use `(princ ...)`  to warn the user, not to send an error and stop the execution.
What do you think about it? Is it better to exit with an error or warn the user, ignore the archive parameter and continue?

test/cask-api-test.el
dgellow: You're talking about `"localhost"` in `(depends-on "package-a" "0.0.1" :archive "localhost")` ?
The doc of `package-pinned-packages` says (emphasis is mine):
> Each element of the alist has the form (PACKAGE . ARCHIVE), where:
    PACKAGE is a symbol representing a package
    ARCHIVE is a **string** representing an archive (it should be the car of
   an element in `package-archives', e.g. "gnu").

dgellow:
@rejeep Can I ask you to review my last changes?

cask.el
```diff
@@ -806,6 +811,14 @@ url to the fetcher source."
         (setf (cask-dependency-ref dependency) ref))
       (-when-let (branch (plist-get args :branch))
         (setf (cask-dependency-branch dependency) branch)))
+    (-when-let (archive (plist-get args :archive))
+      (unless (boundp 'package-pinned-packages)
+        (signal 'cask-no-pinning nil))
+      (setq package-pinned-packages
```
rejeep: The function `add-to-list` is destructive, so you don't need to use `setq`. Also, the third option to `add-to-list` is `append`, you pass the list you are modifying.

rejeep:
Here are a few remarks:

1) Instead of using setting `package-pinned-packages` and using `makunbound`, do this for the case when you want to test that it exist:

```lisp
(let ((package-pinned-packages ...))
  ;; ...
  )
```

And to test what happend when it doesn't exist, mock the `boundp` function, like this:

```lisp
(with-mock
  (mock (boundp 'package-pinned-packages) => nil)
  ;; ...
 )
```

2) I know that package.el wants an archive URL, but the Cask DSL should still support aliases so that you can write this:

```lisp
(source melpa)

(depends-on "foo" :archive melpa)
```

And don't have to write the full URL, like this:

```lisp
(source melpa)

(depends-on "foo" :archive "http://melpa.org/packages/")
```

There is a mapping table you can use to find the URL: https://github.com/cask/cask/blob/f9e804ea96d14e24ae49eeb72d95d44e358b38f2/cask.el#L159

That being said, URLs should still be supported as it might be an archive that Cask does not have an alias for.

dgellow:
Thank you for reviewing my work, I appreciate it! 🙇🏻

About your remarks:

1) 
Something like that?

```elisp
(ert-deftest cask-install-test/pin-archive ()
  (cask-test/with-bundle
   '((source localhost)
     (depends-on "package-a" "0.0.1" :archive "localhost"))
   (let ((package-pinned-packages nil)
         (archive (cask-dependency-archive
                   (car (cask-bundle-runtime-dependencies bundle)))))
     (cask-install bundle)
     (should (equal archive "localhost"))
     (should (equal package-pinned-packages
                    '((package-a . "localhost")))))))
```

The problem is that `cask-add-dependency` modify a global `package-pinned-packages`. If I do a lexical binding via `(let ((package-pinned-packages ;...)))` the test will be done on the local variable, not the global one.
I think that's also the case with `mock`? Not sure about that.

2) Are you sure that `package-pinned-packages` can take a URL? I don't see anything like this in the documentation.
In the context of the test the source is `(source localhost)`, the `:archive "localhost"` isn't a URL but the name of the source, as a string because that's what's needed for `package-pinned-packages`.
I can do a `(symbol-name ;...)` on the archive name so we can specify it as a symbol.

cask.el
dgellow: Ah right, thank you for pointing that out.

rejeep:
1) I see. I'm not very well read up on lexical bindings in Emacs. I wonder if `lexical-let` would help? Or something like this (must be a better way):

```lisp
(let ((old-package-pinned-packages package-pinned-packages))
  (unwind-protect
      (progn
        (setq package-pinned-packages value)
        (should ...))
    (setq package-pinned-packages old-package-pinned-packages)))
```

But as I said, I'm not very well read up on all that. If you don't find a better way, let's leave it as is.

2) Ohh, sorry about that. I forgot that you specify both a name and url. Then I'm fine with the current implementation! 👍

dgellow:
@rejeep What I said before was wrong! `(let ...)` in emacs lisp doesn't do lexical binding. Based on my experience with other lisps I just assumed it did, sorry about that.

I had multiple problems with my use of let:

1) With `let` described as `(let VARLIST BODY)`, the scope in `VARLIST` isn't the same as in `BODY`:

```elisp
(defvar package-pinned-packages) ;; Define global var

(defun just-do-it ()
  (setq package-pinned-packages '(cider . "localhost")) ;; Side effect
  "localhost")

;; Bad
(setq package-pinned-packages "defaultval")

(let ((package-pinned-packages nil)
      (archive (just-do-it)))
  package-pinned-packages) ;; => nil

package-pinned-packages ;; => (cider . "localhost")

;; Good
(setq package-pinned-packages "defaultval")

(let ((package-pinned-packages nil))
  (let ((archive (just-do-it)))
    package-pinned-packages)) ;; => (cider . "localhost")

package-pinned-packages ;; => "defaultval"
```

2) `package-pinned-packages` is set by `cask--eval` when the bundle is created. In the test the creation of the bundle occurs in `(cask-test/with-bundle ...)`.
To override the variable I should do something like this:
```elisp
(ert-deftest cask-install-test/pin-archive ()
  (let ((package-pinned-packages nil)) ;; <= let statement before cask-test/with-bundle
    (cask-test/with-bundle ...)
    (let ...)))
```

dgellow:
But for some reasons my solutions don't work for emacs-23... I will experiment a bit more this weekend and see if I can find what's wrong.

dgellow:
Rebased on current master as my branch was a bit out of date

Fuco1:
Would it be possible to provide multiple archives, so that it tries them in order?

It could further be "automated" by taking the `source` directives and applying them in order, so that the earlier archives are tried first.

dgellow:
@Fuco1 Sure, it can be improved. But my problem at the moment is that I cannot find the reason why my builds fail for emacs-24.3.

I have some free time this week, I can look into it.

PR #357, opened 2016-06-24
Simplify the cask install directory.

phillord:
Previously cask creates a new directory for every version string change,
which, for those running from a development build means most rebuilds.
Now, just use the emacs major and minor version.

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -682,7 +682,10 @@ If BUNDLE is not a package, the error `cask-not-a-package' is signaled."
 
 (defun cask-elpa-path (bundle)
   "Return full path to BUNDLE elpa directory."
-  (f-expand (format ".cask/%s/elpa" emacs-version) (cask-bundle-path bundle)))
+  (f-expand (format ".cask/%s.%s/elpa"
+                    emacs-major-version
+                    emacs-minor-version)
+            (cask-bundle-path bundle)))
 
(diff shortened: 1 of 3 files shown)


rejeep:
@phillord You forgot to update the specs. I don't mind changing to this, what do you say @lunaryorn?

phillord:
Checks pass now....

rejeep:
Great. All I'm concerned about is that this is a breaking change. What do you think, will people realise that it's just a matter of running the install command?

phillord:
Oh, yes, that's true. I hadn't thought about that -- because I develop and rebuild Emacs, this happens to me so often that I don't even notice. I just run `cask install` automatically.

The error messages are better now, I think, so it should report this ("Did you run cask install?").  I could change `cask-elpa-path` to do "set this to emacs-'major-min' unles the directory emacs-'version-number' already exists". That should preserve existing behaviour, at the (significant) cost of making the code more complex and harder to test.

rejeep:
Then I think we should keep it as is. We could bump the minor version to make it a bit more clear that this changes the behaviour.

I'm fine with you merging this now! 👍

phillord:
@fommil I've sent you a starter PR, but I don't know enough about your tests to make it work for sure. Using `cask emacs` should at least set the load path correctly.

PR #359, opened 2016-07-15
Simplify the cask bootstrap directory.

phillord:
Should have done this with the last commit, but hadn't realised that the full version string was used here also.

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -30,9 +30,14 @@
 (eval-when-compile
   (defvar cask-directory))
 
+(defconst cask-bootstrap-emacs-version
+  (format "%s.%s"
+          emacs-major-version
+          emacs-minor-version))
+
 (defconst cask-bootstrap-dir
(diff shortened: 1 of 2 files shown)


rejeep:
👍

rejeep:
@phillord I thought you merged this. I guess this is the cause of https://github.com/cask/cask/issues/368

cask-bootstrap.el
```diff
@@ -30,9 +30,14 @@
 (eval-when-compile
   (defvar cask-directory))
 
+(defconst cask-bootstrap-emacs-version
```
rejeep: Should this perhaps be in `cask.el` called `cask-elpa-dir-version` (or similar). Seems a bit odd that cask.el uses definitions from the bootstrap file (that I see as a self contained file).

rejeep:
Because this is a rather urgent issue I'll merge this PR and perhaps we can change my comment above later on.

PR #363, opened 2016-08-08
Expand cask package target-dir

marsam:
`package-build.el` behaves incorrectly with relative paths.


lunaryorn:
@marsam Thank you :blush:

PR #379, opened 2017-01-12
Use subprocess.call instead of os.execvp for all commands but cask exec

cpitclaudel:
This shouldn't make a significant difference on GNU/Linux or macOS, but
it makes Cask much nicer to use on Windows (http://bugs.python.org/issue9148).

Alternatively, I could make this platform-dependent (subprocess.call on Linux, execvp everywhere else).

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -345,13 +345,13 @@ def exec_cask(arguments):
     Find the Emacs to use for Cask, and run the Cask CLI with the given
     ``arguments``.
 
-    This function replaces the current process.  It does **not** return.
+    This function exits with Emacs' return code.  It does **not** return.
 
     """
     emacs = get_cask_emacs()
     cli = os.path.join(CASK_DIRECTORY, 'cask-cli.el')
     command = [emacs, '-Q', '--script', cli, '--'] + arguments
-    os.execvp(command[0], command)
+    sys.exit(subprocess.call(command))
 
 
 def exit_error(error):


sambrightman:
[This page](https://stackoverflow.com/questions/7264571/how-to-do-os-execv-in-python-in-windows-without-detaching-from-the-console) seems to suggest that using subprocess is non-solution - what problem(s) exactly are occurring on Windows?

cpitclaudel:
IIRC, the problem is that execvp doesn't actually replace the current process on Windows.

sambrightman:
Right, but what problem does this cause for a Cask user?

cpitclaudel:
> Right, but what problem does this cause for a Cask user?

My memories are a bit vague (I only use Windows when debugging Flycheck or esh issues). IIRC, the issue is that as things currently stand `cask commands` run asynchronously on Windows (control is returned to the shell before the command completes, like in bash when you add a `&` to the command.

> subprocess.call won’t replace the current process either...

No (hence the `sys.exit`), but it doesn't return control to the caller.

> Why exclude cask exec?

I think it does matter, on GNU/Linux at least, that `cask exec` uses process replacement (since really it's just setting environment variables and then transferring control to another program).  It could be changed on Windows, but I'm not sure enough of the consequences of doing so to do it.

jcs090218 (approved):

jcs090218:
Thanks for the patch. :)

PR #383, opened 2017-03-08
add package-descriptor directive

### LINKED ISSUE #381: Cask doesn't appear to allow for manually maintained package metadata in its workflow.
Using header comments is a great way to manage dependencies inside of a single file, and Cask supports this workflow in a very straightforward way with the `package-file` directive. However, if your package has a lot of dependencies, then it starts to become unwieldly. The dependency header must appear on a single line and so with a configuration package like we're developing that has around 40 dependencies, it's untenable (I haven't found a way to make them multi-line)

For that reason, I wanted to manage dependencies in elisp. My first attempt was to invoke the `package-file` directive with the package descriptor:

```elisp
(package-file "my-pkg.el")
```
but that assumed it was a package file, not a descriptor. Failing that, I used a combination of `package` and `depends-on`:

```elisp
(package "my")
(depends-on "foo")
(depends-on "bar")
```
This worked great in development, but when it came time to deploy the package, the `my-pkg.el` generated by `cask package` did not actually include `foo` and `bar` in its list of dependencies.

This makes it difficult to work with manually maintained multi-file package metadata. Some possible remedies to this are:

1. make the `package-file` directive "smart" so that if it ends in `-pkg.el`, then it treats it as a package descriptor
2. Add a new directive: `package-descriptor` that assumes that allows you to incorporate a package descriptor directly.
3. If you use a combination of `package` and `depends-on` directives, then make sure the resulting package descriptor generated by `cask package` contains those dependencies.

I'd be happy to take a stab at adding that if I can get some guidance as to the best option. I'm leaning towards (2) since it seems like the most intuitive.

cowboyd:
Sometimes you want to just explicitly enumerate package metadata yourself in a `-pkg.el` that contains everything from the name, to the dependencies.

This allows Cask to read package metadata from a pre-existing package descriptor by adding the `package-descriptor` directive. This directive functions in a very similar fashion to the `package-file` directive. For example, if you have a package descriptor in `my-package-pkg.el"`, that looks like:

```elisp
(define-package "my-package" "1.0.0" "a great package"
  '((foo "2.8.0")
    (bar "0.7.0")
    (bazz "0.13.0"))
  :keywords '("awesome" "possum"))
```

You can have Cask use it by including this line in your `Cask` file.

```elisp
(package-descriptor "my-package-pkg.el")
````
resolves #381 

## TODOS

- [x] Find out where is the appropriate place to test
- [x] Customize any errors where appropriate
- [x] Add documentation

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -393,13 +393,17 @@ SCOPE may be nil or 'development."
                      (cask--show-package-error err filename)))))
              (cask--from-epl-package bundle package))))
         (package-descriptor
-         (cl-destructuring-bind (_ filename) form
-           (let ((package
-                  (condition-case err
+         (cl-destructuring-bind (_ &optional filename) form
+           (let* ((descriptor-filename
+                   (or filename (let ((pkg-files (f-glob "*-pkg.el" (s-chop-suffix "/" (cask-bundle-path bundle)))))
+                                  (if (car pkg-files) (f-filename (car pkg-files))
+                                    (error "No -pkg.el file found for package descriptor")))))
+                  (package
+                   (condition-case err
                       (epl-package-from-descriptor-file
-                       (f-expand filename (cask-bundle-path bundle)))
+                       (f-expand descriptor-filename (cask-bundle-path bundle)))
                     (epl-invalid-package
-                     (cask--show-package-error err filename)))))
+                     (cask--show-package-error err descriptor-filename)))))
              (cask--from-epl-package bundle package))))
         (depends-on
          (cl-destructuring-bind (_ name &rest args) form
--- a/cask.el
+++ b/cask.el
@@ -392,6 +392,15 @@ SCOPE may be nil or 'development."
                     (epl-invalid-package
                      (cask--show-package-error err filename)))))
              (cask--from-epl-package bundle package))))
+        (package-descriptor
+         (cl-destructuring-bind (_ filename) form
+           (let ((package
+                  (condition-case err
+                      (epl-package-from-descriptor-file
+                       (f-expand filename (cask-bundle-path bundle)))
+                    (epl-invalid-package
+                     (cask--show-package-error err filename)))))
+             (cask--from-epl-package bundle package))))
         (depends-on
          (cl-destructuring-bind (_ name &rest args) form
            (when (stringp (car args))
--- a/cask.el
+++ b/cask.el
@@ -156,6 +156,16 @@ Slots:
 `column' Column number on the `line'."
   line column)
 
+;; Specializations of package-build classes and methods to define a
+;; directory based recipe.
+(defclass package-directory-recipe (package-recipe)
+  ((dir           :initarg :dir   :initform ".")))
+
+(defmethod package-recipe--working-tree ((rcp package-directory-recipe))
+  (oref rcp dir))
+
+(defmethod package-build--get-commit ((rcp package-directory-recipe)))
+
 (defvar cask-source-mapping
   '((gnu          . "https://elpa.gnu.org/packages/")
     (melpa        . "https://melpa.org/packages/")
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by cowboyd** [comment]
@rejeep @lunaryorn Let me know what you think. 

I wasn't sure where the best place to test this was, and what areas of the documentation should be updated to encompass this feature, so I left them as TODOS. Any guidance you could provide would be most helpful.

**Comment by lunaryorn** [comment]
@cowboyd I do not contribute to Cask anymore, so I'm unsubscribing from this discussion.  Please do not mention me for Cask issues.

**Comment by cowboyd** [comment]
What about you @rejeep? You're the only other person listed in the organization. 

Your thoughts? I'm particularly timid about where the best place to add a test is.

**Comment by cowboyd** [comment]
@sambrightman @davidshepherd7 

Any thoughts on this? We've been dependent on this feature for awhile and would love to see it get merged up into master. If this doesn't make sense, then would love to the best way forward.

The reason that we need this feature is that we have a very large package that we maintain with a lot of dependencies; so many in fact, that they just don't fit inside of the comments of the main elisp package 

https://github.com/thefrontside/frontmacs/blob/master/frontmacs-pkg.el

**Comment by cowboyd** [comment]
ping @rejeep

**Comment by sambrightman** [comment]
I don't have commit rights, so I can't help you merge this. Why is it better than option 3 in #381?

**Comment by cowboyd** [comment]
I went for option 2 because it turned out to be the easier to implement, but really I don't have a preference. Certainly it was a surprise when the generated `-pkg.el` file ignored all of my `depends-on` directives.

I was hoping to kickstart a discussion around this, but haven't been able to get a response from anybody.

**Comment by cowboyd** [comment]
@sambrightman Which option would you prefer and why?

**Comment by rejeep** [comment]
I'm not involved in the project anymore so I don't really have an opinion, but if someone need commit access or if you need me to merge something when all agree on a solution I can help out.

**Comment by sambrightman** [comment]
@cowboyd I don't have a strong opinion because I don't know the issue well. 3) seems like a bug, so fixing that concomitantly with solving your problem seems better to me than adding to the API.

@rejeep I very much doubt I can be a full maintainer but would like commit access to get one or two issues moving.

**Comment by rejeep** [comment]
@sambrightman Sent you and invite!

**Comment by sambrightman** [comment]
Given the lack of response about `package-build` functionality, let's move forward with this. I think the documentation should highlight the rationale/difference from `package-file`.  I believe you want to add   a fixture and tests for `cask-setup`.

Does it make sense to allow no file specification if there is only one file matching `*-pkg.el`?

**Comment by cowboyd** [comment]
Thanks Sam. I'll go ahead and get moving on this.

> Does it make sense to allow no file specification if there is only one file matching *-pkg.el?

I'm not sure I understand completely. Are you asking if your package contains literally one file called *-pkg.el then we infer that it is the package descriptor?

**Comment by sambrightman** [comment]
Yes; I would assume that it is common to have only one such file, and this would avoid needing to name it (thus, not break on renames either). No strong preference though.

**Comment by sambrightman** [comment]
Did you make any progress on this?

**Comment by cowboyd** [comment]
I haven't made any progress on this yet 😄 However, I'm hoping to dig into it this week.

**Comment by cowboyd** [comment]
@sambrightman Ok, I took a stab at writing the feature specification for this https://github.com/cowboyd/cask/blob/2d89062ab548e3a27a216592334221bfb9c38a09/features/package-descriptor.feature

Do you think that these are valid acceptance criteria to drive for? If not, let me know and I'll adjust. If so, let me know and I'll get started 😄

**Comment by cowboyd** [comment]
@sambrightman First pass is ready. This feature is implemented, tested and now, if you just have a 

`(package-descriptor)` directive directly in your `Cask` file, then it will pull in the first `-pkg.el` file that it finds.

Let me know what you think!

**Comment by cowboyd** [comment]
@sambrightman Ping.

(fixed failing tests)

**Comment by cowboyd** [comment]
@sambrightman We should be ready to go here. If there are any other changes, please let me know.

File: cask.el [hunk 1]
```diff
@@ -392,6 +412,19 @@ SCOPE may be nil or 'development."
                     (epl-invalid-package
                      (cask--show-package-error err filename)))))
              (cask--from-epl-package bundle package))))
+        (package-descriptor
+         (cl-destructuring-bind (_ &optional filename) form
+           (let* ((descriptor-filename
+                   (or filename (let ((pkg-files (f-glob "*-pkg.el" (s-chop-suffix "/" (cask-bundle-path bundle)))))
```
sambrightman: `s-chop-suffix` seems unnecessary here. If it were necessary, it seems more likely that `directory-file-name` should be used.

sambrightman: I've made this change locally and will update the PR, but having trouble getting tests going on Emacs 26. I'll merge like this and if you have trouble on Emacs > 25 we can look at it afterwards.

sambrightman:
Thanks for the contribution and patience in bringing this to fruition!

PR #384, opened 2017-03-14
Use files pattern similar to package-build

marsam:
Hi:

I imported the file list used [`package-build.el`](https://github.com/melpa/package-build/blob/9a6824c/package-build.el#L1013-L1023).

This solves the scenario of when a package uses the same recipe as melpa. e.g:

``` emacs-lisp
(package-file "mypkg.el")

(files :defaults)
```

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -770,8 +770,15 @@ If no such dependency exist, return nil."
 This is done by expanding the patterns in the BUNDLE path.  Files
 in the list are relative to the path."
   (cask--with-file bundle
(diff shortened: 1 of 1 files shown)


sambrightman:
Could you add `cask-file` tests for this (`:defaults alone`, appending to it, etc.)?

marsam:
Sure!
btw, are you the new maintainer? I'll update it if there is chance of this being merged.

sambrightman:
I'm not officially maintaining but trying to keep things moving. I will merge after you update.

PR #385, opened 2017-04-02
cask init: autofill the package-file value

Wilfred:
If we're in a directory that only contains a single foo.el, that's
probably the value we want for package-file.

Before:

    (package-file "TODO")

After:

    (package-file "foo.el")

If it's ambiguous, use the previous behaviour.

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -632,7 +632,7 @@ configuration template is used."
     ;; If there's only a single .el file, use that as the package-file.
     (when dev-mode
       (let* ((files (f-files (cask-path bundle)))
-             (el-files (--filter (s-ends-with? ".el" it) files))
+             (el-files (--filter (f-ext? it "el") files))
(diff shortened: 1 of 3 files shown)


cask.el
```diff
@@ -629,6 +629,14 @@ configuration template is used."
   (let ((cask-file (cask-file bundle))
         (init-content
          (cask--template-get (if dev-mode "init-dev.tpl" "init.tpl"))))
+    ;; If there's only a single .el file, use that as the package-file.
+    (when dev-mode
+      (let* ((files (f-files (cask-path bundle)))
+             (el-files (--filter (s-ends-with? ".el" it) files))
```
sambrightman: `f-ext?` seems better here.

sambrightman:
Could you add `cask-caskify` tests?

PR #388, opened 2017-05-09
Do not show the optional arg of --help in usage

snogge:
The --help option should only be used to display the usage
information.  This is the standard way to define command line options.


sambrightman:
Like many tools, Cask does offer command-specific help using the syntax described (although it is just a subset of the main usage information). I don't see any reason not to show this in the usage.

PR #392, opened 2017-07-05
Use only emacs-version in cask-bootstrap-emacs-version

shigemk2:
- GNU Emacs 25.2.1
- Mac OSX 10.12.4

Emacs cannot load packages because current cask downloads packages as "25.2".
It is because cask downloads Emacs packages emacs-major-version + emacs-minor-version as "25.2".

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -31,9 +31,9 @@
   (defvar cask-directory))
 
 (defconst cask-bootstrap-emacs-version
-  (format "%s.%s"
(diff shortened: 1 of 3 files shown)


shigemk2:
@rejeep 
Could you review this pull-request?

sambrightman:
You need to run `cask install` when changing Emacs versions (this is intentional behaviour, see http://cask.readthedocs.io/en/latest/guide/usage.html#finding-emacs).

PR #400, opened 2017-09-29
If /usr/bin is at the front of $PATH, strip it

### LINKED ISSUE #399: Cask recently started setting the wrong $PATH on Travis
My package has been using evm and cask without issue on Travis for a little over a month. But apparently something changed in the past few weeks that causes cask to ignore the evm-installed emacs version. The reason seems to be that cask is adding `/usr/bin/` to the front of `$PATH`, which means that the system `/usr/bin/emacs` takes precedence over the evm copy of Emacs. This is demonstrated in the latest Travis build log: https://travis-ci.org/DarwinAwardWinner/ido-completing-read-plus/jobs/277499771

In that build, I print out `$PATH` with and without cask:

```
$  echo $PATH
/home/travis/.cask/bin:/home/travis/.evm/bin:...
$ cask exec sh -c 'echo $PATH'
/usr/bin:/opt/pyenv/libexec:/opt/pyenv/plugins/python-build/bin:/home/travis/.cask/bin:/home/travis/.evm/bin:...
```

I'm not sure what cask was adding to $PATH back when my build used to work, since I only added that debug code once the build failures started.

Unfortunately, there also appears to be a separate issue that prevents cask's `--verbose` and `--debug` options from working, so I was unable to add them to the cask invocations in my `.travis.yml`. Specifically, if the first argument to cask starts with a dash, cask insists on interpreting *every* argument that begins with a dash anywhere in the command. Example:

```
$ cask exec sh -c 'echo hello'
hello
$ cask --verbose exec sh -c 'echo hello'
Option ‘-c‘ not available
$ cask --debug exec sh -c 'echo hello'
Debugger entered--Lisp error: (error "Option ‘-c‘ not available")
  signal(error ("Option ‘-c‘ not available"))
  error("Option `%s` not available" "-c")
  commander--handle-options(("--debug" "exec" "sh" "-c" "echo hello"))
  commander-parse(("--debug" "exec" "sh" "-c" "echo hello"))
  (if commander-parsing-done nil (commander-parse (or commander-args (cdr command-line-args-left))))
  eval-buffer(#<buffer  *load*> nil "/home/ryan/.cask/cask-cli.el" nil t)  ; Reading at buffer position 12803
  load-with-code-conversion("/home/ryan/.cask/cask-cli.el" "/home/ryan/.cask/cask-cli.el" nil t)
  load("/home/ryan/.cask/cask-cli.el" nil t t)
  command-line-1(("-scriptload" "/home/ryan/.cask/cask-cli.el" "--" "--debug" "exec" "sh" "-c" "echo hello"))
  command-line()
  normal-top-level()
```

davidshepherd7:
This is a workaround for a bug in pyenv which is affecting Travis
builds. pyenv is currently adding `/usr/bin` to the front of $PATH for
anything run through python. This screws up emacs version management
with evm which relies on the evm bin directory being earlier in the path
than any other directory containing emacs binaries.

Fixes #399.

### DIFF
--- a/features/exec.feature
+++ b/features/exec.feature
@@ -64,3 +64,22 @@ Feature: Exec
       cask exec: error: Failed to execute does-not-exist: [Errno 2] No such file or directory
       Did you run cask install?
       """
+
+  Scenario: With Travis' pyenv prepending /usr/bin to path, remove the first instance
+    Given this Cask file:
+      """
+      """
+    When I set environment variable "TRAVIS" to "true"
+    When "/usr/bin" is prepended to $PATH
+    When I run cask "exec sh -c 'echo $PATH'"
+    Then I should not see command output matching "^/usr/bin:"
+    Then I should see command output matching ":/usr/bin:"
+
+  Scenario: With a path starting with /usr/bin, but not on Travis
+    Given this Cask file:
+      """
+      """
+    When I set environment variable "TRAVIS" to "blah"
+    When "/usr/bin" is prepended to $PATH
+    When I run cask "exec sh -c 'echo $PATH'"
+    Then I should see command output matching "^/usr/bin:"
--- a/bin/cask
+++ b/bin/cask
@@ -317,6 +317,13 @@ def exec_command(command):
     if (b'CASK_EMACS' in ENVB):
         ENVB[b'EMACS'] = ENVB[b'CASK_EMACS']
 
+    # Special handling for emacs on travis with evm and buggy pyenv, see cask
+    # issue #399.
+    if ENVB.get(b'TRAVIS', b'') == b'true':
+        paths = ENVB[b'PATH'].split(b':')
+        if len(paths) > 0 and paths[0] == b'/usr/bin':
+            ENVB[b'PATH'] = ':'.join(paths[1:])
+
     try:
         os.execvp(command[0], command)
     except OSError as error:
--- a/features/step-definitions/cask-steps.el
+++ b/features/step-definitions/cask-steps.el
@@ -72,6 +72,14 @@
                 (t
                  (setq cask-test/stderr content))))))))
 
+(When "^\"\\([^\"]+\\)\" is prepended to $PATH$"
+  (lambda (path)
+    (setenv "PATH" (s-concat path ":" (getenv "PATH")))))
+
+(When "I set environment variable \"\\([^\"]+\\)\" to \"\\([^\"]+\\)\""
+  (lambda (var value)
+    (setenv var value)))
+
 (Then "^I should see command error:$"
   (lambda (output)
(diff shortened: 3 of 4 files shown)


davidshepherd7:
The Travis checks are not really reliable because of the bug that this PR fixes.

@rejeep I know you are busy these days but it would be great if when you do have some time you could prioritize this PR because at the moment pretty much all Travis emacs testing with cask and evm is broken! In particular I can't easily test my fix for ecukes on all emacs versions because of this bug :disappointed:.

rejeep:
@davidshepherd7 Don't you want `/usr/bin` on `PATH`? Why not move it to the end?

davidshepherd7:
The problem with pyenv is that it adds a second copy of `/usr/bin` to the front of the path, so we don't want that second copy there at all. 

Now that you've pointed it out though I've realised this change could cause problems if someone happens to have `/usr/bin` at the front of their path for some other reason. I think the best thing to do would be to add something like `ENVB[b'TRAVIS'] == b'true'` to the if statement, that way we won't mess up anyone's use of cask outside Travis and we can be sure that the path will be as expected.

P.S. sorry for not replying sooner, for some reason my phone is not giving me notifications for new emails.

rejeep:
@davidshepherd7 I'm not really a fan of "Cask knowing what Travis is". Both because I don't Cask should know that and also that this issue would only be solved on Travis. I would much rather check if pyenv is used. Would that work?

davidshepherd7:
@rejeep Yeah I agree it's not ideal, but pyenv doesn't seem to document any environment variable that it sets which we could use to check if it's active or not. We could check for the presence of `.*/shims` in `PATH`, but I think that's likely to give false positives.

Also I think it's only the specific pyenv setup on the default Travis container that causes the problem. As far as I can tell in normal use it prepends some other directory (containing binaries for the python version you've selected) instead of `/usr/bin`. So I think it's less likely that people will run into this issue outside of Travis.

features/exec.feature
```diff
@@ -64,3 +64,22 @@ Feature: Exec
...
       Did you run cask install?
       """
+
+  Scenario: With Travis' pyenv prepending /usr/bin to path, remove the first instance
+    Given this Cask file:
+      """
+      """
+    When I set environment variable "TRAVIS" to "true"
+    When "/usr/bin" is prepended to $PATH
+    When I run cask "exec sh -c 'echo $PATH'"
```
rejeep: Isn't this the same as `When I run cask "exec-path"`? (same below)

rejeep:
@davidshepherd7 Ok, too bad. I just had a minor comment. I'm fine with the changes as it seems there is no other option. I added you as a collaborator, so feel free to merge.

PR #403, opened 2017-10-13
Fix VERSION_RE

vapniks:
Fix VERSION_RE to work with emacs running in firejail


sambrightman:
I don't know firejail - why does the existing regex not work?

vapniks:
firejail is a sandboxing security program: https://firejail.wordpress.com/
I have emacs (and other apps) symlinked to firejail which means it outputs some firejail messages on the command line before the emacs messages.

vapniks:
actually, just discovered it's possible to suppress the firejail output using the "quiet" command.

sambrightman:
That sounds better, great!

PR #404, opened 2017-10-15
If /usr/bin is at the front of $PATH, strip it -- attempt 2

### LINKED ISSUE #399: Cask recently started setting the wrong $PATH on Travis

davidshepherd7:
This is a second attempt at fixing this issue, the first one didn't work in some cases so I rolled it back. This time I managed to test my changes on Travis (by messing with the bootstrap process), and it does pick the correct emacs version as you can see [here](https://travis-ci.org/davidshepherd7/electric-operator/jobs/288197732).

The only thing I've changed from the previous attempt is that the PATH modifications are now in `main` instead of only in `exec_command` (which isn't always used). So now the path is fixed in everything that cask does instead of only when using `cask exec ...` or `cask emacs ...`.

---------------------------------------------------------------------------------------------------------

This is a workaround for a bug in pyenv which is affecting Travis
builds. pyenv is currently adding `/usr/bin` to the front of $PATH for
anything run through python. This screws up emacs version management
with evm which relies on the evm bin directory being earlier in the path
than any other directory containing emacs binaries.

Fixes #399.

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -379,6 +379,14 @@ def main():
 
     """
     try:
+        # Special handling for emacs on travis with evm and buggy pyenv, see
+        # cask issue #399.
+        if ENVB.get(b'TRAVIS', b'') == b'true':
+            paths = ENVB[b'PATH'].split(b':')
+            if len(paths) > 0 and paths[0] == b'/usr/bin':
+                ENVB[b'PATH'] = ':'.join(paths[1:])
+
+
         if sys.version_info[:2] < (2, 6):
             exit_error(
                 'Python 2.6 required, yours is {0.major}.{0.minor}'.format(
--- a/features/exec.feature
+++ b/features/exec.feature
@@ -64,3 +64,22 @@ Feature: Exec
       cask exec: error: Failed to execute does-not-exist: [Errno 2] No such file or directory
       Did you run cask install?
       """
+
+  Scenario: With Travis' pyenv prepending /usr/bin to path, remove the first instance
+    Given this Cask file:
+      """
+      """
+    When I set environment variable "TRAVIS" to "true"
+    When "/usr/bin" is prepended to $PATH
+    When I run cask "exec sh -c 'echo $PATH'"
+    Then I should not see command output matching "^/usr/bin:"
+    Then I should see command output matching ":/usr/bin:"
+
+  Scenario: With a path starting with /usr/bin, but not on Travis
+    Given this Cask file:
+      """
+      """
+    When I set environment variable "TRAVIS" to "blah"
+    When "/usr/bin" is prepended to $PATH
+    When I run cask "exec sh -c 'echo $PATH'"
+    Then I should see command output matching "^/usr/bin:"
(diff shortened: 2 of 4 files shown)


davidshepherd7:
@rejeep Ping :slightly_smiling_face:

It's basically the same as the previous PR (which didn't quite work in some cases) but I'd appreciate you having a quick look at it in case I've done something stupid!

davidshepherd7:
@rejeep I'm just going to merge this because it's very similar to a previous PR that you approved and it's  causing problems elsewhere. If you like you can review it when you have time and I can fix any issues that come up then.

Hope that's ok :+1:

rejeep:
@davidshepherd7 Sorry for not answering, but I don't have enough time so I lately I have ended up ignoring all message. I'm fine with you merging this, you seem to know what you are doing. Hope it works out for you.

davidshepherd7:
@rejeep Don't worry about it, life comes first! :+1:

PR #408, opened 2017-12-12
convert package-name to string

### LINKED ISSUE #407: Incompatibility with package-build-package
I am not 100% sure if this is a Cask bug or a package-build bug:

When using the VCS version of `depends-on`, package-build.el throws a:

```
Dependency <package-name> failed to install: Wrong type argument: sequencep, <package-name>
```

I tracked the error to package-build.el's [`package-build--build-multi-file-package`](https://github.com/melpa/package-build/blob/master/package-build.el#L1075), where `package-name` arrives as a symbol, but apparently is expected as a string.

On cask's side of things, `depends-on` transform package names from strings to symbols [here](https://github.com/cask/cask/blob/master/cask.el#L400), but I think it has been this way since long time ago.

jvshahid:
fixes #407

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -320,7 +320,7 @@ This function returns the path to the package file."
          (config (cask--dependency-to-package-build-config dependency))
          (files (cask--dependency-files dependency)))
     (let ((version (package-build-checkout name config path)))
-      (package-build-package name version files path cask-tmp-packages-path)
+      (package-build-package (symbol-name name) version files path cask-tmp-packages-path)
       (let ((pattern (format "%s-%s.*" name version)))
         (car (f-glob pattern cask-tmp-packages-path))))))
 


jvshahid:
ping. this is really annoying. upgrading cask makes it unusable. I'm happy to make additional changes or clean up the pr. thanks!

sambrightman:
Per #407, this is caused by https://github.com/melpa/package-build/commit/063e64ad55ac07a348af0373605fd90348a8bc11. It would be good to have tests and assess backward-compatibility. I get the impression that doing `cask upgrade-cask` would be necessary to see either the cause or the solution, so most people will be okay after this change. People who already have the problem should be able to fix it with `cask upgrade-cask`. Have you tested some scenarios?

jvshahid:
@sambrightman sorry, but I am not sure I understand your question, specially the `cask upgrade-cask` part. I can't get the tests to fail prior to my commit or with my commit. I think we are lacking test coverage for the interaction with `package-build`.

sambrightman:
I'm suggesting that this PR include the missing tests and asking that we gain some confidence of not introducing breakage for anyone else.

jvshahid:
@sambrightman it makes sense to add a test but i'm not sure what value are we getting from ensuring backward compatability ? wouldn't `cask upgrade` upgrade cask as well as its dependencies. how can someone end up with a newer version of cask and an older version of package-build ?

jvshahid:
That said, I am definitely happy to add some test cases for this pr

jvshahid:
it turns out i was wrong. Without this change the unit test suite is currently failing. I am getting the following failures:

```
Ran 126 tests in 69.564 seconds
5 unexpected results:
   FAILED  cask-install-test/fetcher-branch
   FAILED  cask-install-test/fetcher-files
   FAILED  cask-install-test/fetcher-git
   FAILED  cask-install-test/fetcher-install-twice
   FAILED  cask-install-test/fetcher-ref
```

They all fail with the same error:

````
    (cask-failed-installation
     [cl-struct-cask-dependency package-a nil :git "/tmp/cask32347A7G/" nil "472b2ffd79a52a9c1970cf97a7e7dc2eaf918644" nil]
     (wrong-type-argument sequencep package-a)
     "")
```

The only reason i think travis is green is because the tests didn't run in 2 months. you should be able to validate by triggering travis to run the test suite again. With that said, what else would you like to see in this pr to merge it in.

sambrightman:
Yes, confirmed.

If you re-read my comment, I'm not suggesting adding backward-compatibility. I'm suggesting we assess any scenarios where new breakage might occur - maybe for more people than are currently broken. I don't think the ways in which new breakage can occur are worse than the current situation though, so better just to fix.

PR #415, opened 2018-02-13
Fix os.exec in Windows

chrycheng:
Using `os.exec*` causes cask to return to prompt while the command is still running.  Python documentation does not recommend setting `shell` to `True` and even claims it isn't necessary to run a batch file.  However, I didn't find the latter to be the case during testing.  The behavior of `cask exec` is slightly different in Windows as a result.  No `OSError` is thrown when the command does not exist.


sambrightman:
This was previously brought up as #379. I looked over that PR a little but:

1. The problem wasn't clear to me.
2. It wasn't clear to me if the PR solved the problem.
3. There was concern about possible changes to non-Windows versions.
4. It seemed important that `exec` remained the same on Linux (but not why).

I think 1) is about the async behaviour - a user would have to manually wait for completion of Cask? The others are still not 100% clear, although I assume you have tested 2).

chrycheng:
Sorry, I didn't think to check any open PRs.  This one fixes one more feature and I think is more robust.

> I think 1) is about the async behaviour - a user would have to manually wait for completion of Cask?

That's correct.  As [documented](https://docs.python.org/2/library/os.html#os.execl), in Unix, `exec*` replaces the calling program with the specified program.  This isn't the case in Windows where the specified program is run in a separate independent process.  The calling program doesn't wait for the other one so, when it finishes, the user is brought back to the command prompt while the other program is still executing in the background.  Here is a demo:

cask-pr-415-demo

> The others are still not 100% clear, although I assume you have tested 2).

Yes, I have tested that the PR solved the problem in Windows.  As far as I can tell, the Linux behavior remained the same.  However, I can push another commit that limits the change to Windows only if that is what you would prefer.  I thought about doing this at first but I thought it might be bad practice.

jcs090218:
I have temporary merged #379. Think PR is indeed more robust. I will leave this PR open just in case I feel like these pre-check are necessary. :)

jcs090218 (approved):

PR #416, opened 2018-02-14
Support VT codes in Windows

chrycheng:


sambrightman:
Can you provide some context on what this PR is trying to achieve? Is this for coloured output from tests?

chrycheng:
Sorry, yes, that's correct.  It enables the [Windows console to interpret virtual terminal sequences](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences).  Without it, they would appear as plain characters on screen like so:

uninterpreted-vt-codes

They should appear colored like so:

interpreted-vt-codes

Fuco1:
I don't see any problem with merging this.  WDYT?  Worst case we break the 3 windows users who run Emacs there :)

jcs090218:
Does this appear in powershell? If this works in powershell then I might pass this PR.

PR #423, opened 2018-04-11
Adjust cask to new package-build APIs

### LINKED ISSUE #421: 'cask package' is broken
A recent refactoring of [melpa/package-build](https://github.com/melpa/package-build/) broke `cask package`.

AFAICT the signature of `package-build-package` changed signature which is causing errors like these when running `cask package`:

```
Wrong number of arguments: #[(rcp version) "!�!�È	
\"�Ä\"�
ÊË$��(ÌÍ\"����4ÎÏ\"�[�GÐU�G���@@	$�[Ð�GW�X���	$�[ÎÓ!,�" [rcp source-dir file-specs files name package-build-default-files-spec package-recipe--working-tree package-build--config-file-list package-build-expand-file-specs eieio-oref ...] 6 ("/home/travis/.emacs.d/.cask/25.3/bootstrap/package-build-20180330.2032/package-build.elc" . 30212)], 5
make: *** [package-test] Error 255
```

I have seen similar errors for emacs versions between 24.5 to 25.3

dajva:
- Adjust cask-package to use package-recipe.
- Use package-recipe when interacting with package-build--checkout

Fixes #421

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -41,7 +41,7 @@
   "Path to Cask bootstrap directory.")
 
 (defconst cask-bootstrap-packages
-  '(s dash f commander git epl shut-up cl-lib package-build)
+  '(s dash f commander git epl shut-up cl-lib package-build eieio)
   "List of bootstrap packages required by this file.")
 
 (unless (require 'package nil :noerror)
--- a/cask.el
+++ b/cask.el
@@ -156,6 +156,16 @@ Slots:
 `column' Column number on the `line'."
   line column)
 
+;; Specializations of package-build classes and methods to define a
+;; directory based recipe.
+(defclass package-directory-recipe (package-recipe)
+  ((dir           :initarg :dir   :initform ".")))
+
+(defmethod package-recipe--working-tree ((rcp package-directory-recipe))
+  (oref rcp dir))
+
+(defmethod package-build--get-commit ((rcp package-directory-recipe)))
+
 (defvar cask-source-mapping
   '((gnu          . "https://elpa.gnu.org/packages/")
     (melpa        . "https://melpa.org/packages/")
@@ -300,14 +310,23 @@ ARGS is a plist with these additional options:
   "Return true if BUNDLE has any fetcher dependencies."
   (> (length (cask--fetcher-dependencies bundle)) 0))
 
-(defun cask--dependency-to-package-build-config (dependency)
-  "Turn DEPENDENCY into a package-build config object."
-  (let ((fetcher (intern (substring (symbol-name (cask-dependency-fetcher dependency)) 1)))
+(defun cask--dependency-to-package-build-recipe (dependency)
+  "Turn DEPENDENCY into a package-build recipe object."
+  (let ((name (symbol-name (cask-dependency-name dependency)))
         (url (cask-dependency-url dependency))
         (files (cask--dependency-files dependency))
+        (fetcher (substring (symbol-name (cask-dependency-fetcher dependency)) 1))
         (commit (cask-dependency-ref dependency))
         (branch (cask-dependency-branch dependency)))
-    (list :fetcher fetcher :url url :files files :commit commit :branch branch)))
+    (cask--create-package-build-recipe
+     fetcher name
+     :url url :files files :commit commit :branch branch)))
+
+(defun cask--create-package-build-recipe (fetcher name &rest args)
+  "Create a package-build `package-recipe' for FETCHER with NAME.
+ARGS is the initialization slots."
+  (let ((constructor (intern (format "package-%s-recipe" fetcher))))
+    (apply constructor name :name name args)))
 
 (defun cask--checkout-and-package-dependency (dependency)
   "Checkout and package DEPENDENCY.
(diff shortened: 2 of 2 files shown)


cask.el
```diff
@@ -156,6 +156,16 @@ Slots:
 `column' Column number on the `line'."
   line column)
 
+;; Specializations of package-build classes and methods to define a
```
dajva: This stuff should ideally go into https://github.com/melpa/package-build to avoid having it separately. The failing travis builds for emacs 24.3 could possibly be solved by moving the code there. Just a wild guess really since I don't know `eieio` at all.
@tarsius: What do you say about this? Does it make sense to suppor directory based packages in `package-build`?

cask.el
```diff
@@ -300,29 +310,39 @@ ARGS is a plist with these additional options:
...
-         (files (cask--dependency-files dependency)))
-    (let ((version (package-build-checkout name config path)))
-      (package-build-package (symbol-name name) version files path cask-tmp-packages-path)
+  (let ((name (symbol-name (cask-dependency-name dependency)))
+        (rcp (cask--dependency-to-package-build-recipe dependency))
+        (package-build-working-dir cask-tmp-checkout-path)
+        (package-build-archive-dir cask-tmp-packages-path) )
+    (let ((version (package-build--checkout rcp)))
+      (package-build-package rcp version)
       (let ((pattern (format "%s-%s.*" name version)))
```
tarsius: Maybe you could combine these three `let`s into a single `let*`? If so it would be best to do that in a separate commit.

tarsius:
@dajva said:

> This stuff should ideally go into https://github.com/melpa/package-build to avoid having it separately. [...] @tarsius: What do you say about this? Does it make sense to support directory based packages in package-build?

(We probably don't want this to disappear in "Obsolete changes".)

Well, first of all thanks and I am glad that all it takes to make Cask compatible again is a bit of glue code.

It might make sense to add this to `package-build` itself again, but I think it would be better to keep it separate for the time being. I would suggest that you put all the "wrap `package-build` so that it behaves we think it should be have" as well as the definitions of additional fetcher classes into a separate library. If nothing else that would it make it easier for the people who work on `package-build` to see how it is being used by Cask without having to dig through lots of unrelated code.

My bold changes to `package-build` have caused some issues for downstreams, but I think the changes are quite necessary. There are a few projects built in top of `package-build` but the maintainers of these projects have not contributed back much in the past. I think it would have been better to - more slowly than I have done now - improve `package-build` than to just accept it as-is. I hope that the recent breakage means that downstreams will now at least pay more attention to changes proposed by others.

I have little time to work on this again soon, but I think that due to my changes it is now much easier to read `package-build` and learn what it does and how, possibly for the purpose of improving that.

tarsius:
Ps. For people interested in improving `package-build`: You might want to also look at package building support in `package.el` itself as well as in utilities used by GNU Elpa (and tracked in the same repository as the packages). These tools probably probably do some things better than `package-build.el`, others worse. It would be nice if we could port the good parts and eventually even arrive at a single package providing at least all the functionality required by all downstreams.

dajva:
@sambrightman: The fixup for emacs 24.3 supposedly fixes the failing test issues with this patch. I don't have the power to restart the build though. Maybe you can do it?

cowboyd:
@dajva Maybe with an empty commit, or an amended commit?

dajva:
@cowboyd: Thanks for the tip. Worked perfectly!

dajva:
@sambrightman: Tests are passing. Could you have a look at this pr?

sambrightman:
Sorry for the long delay, I'm looking at this now.

cask-bootstrap.el
```diff
@@ -41,7 +41,7 @@
   "Path to Cask bootstrap directory.")
 
 (defconst cask-bootstrap-packages
-  '(s dash f commander git epl shut-up cl-lib package-build)
+  '(s dash f commander git epl shut-up cl-lib package-build eieio)
```
sambrightman: I don't think built-in packages need adding here?

cask.el
```diff
@@ -300,29 +310,39 @@ ARGS is a plist with these additional options:
...
-         (path (f-expand (symbol-name name) cask-tmp-checkout-path))
-         (config (cask--dependency-to-package-build-config dependency))
-         (files (cask--dependency-files dependency)))
-    (let ((version (package-build-checkout name config path)))
-      (package-build-package (symbol-name name) version files path cask-tmp-packages-path)
+  (let ((name (symbol-name (cask-dependency-name dependency)))
+        (rcp (cask--dependency-to-package-build-recipe dependency))
+        (package-build-working-dir cask-tmp-checkout-path)
+        (package-build-archive-dir cask-tmp-packages-path) )
+    (let ((version (package-build--checkout rcp)))
```
sambrightman: This now depends on a private method - is there a better way?

cask.el
```diff
@@ -300,29 +310,39 @@ ARGS is a plist with these additional options:
...
+      (package-build-package rcp version)
       (let ((pattern (format "%s-%s.*" name version)))
-        (car (f-glob pattern cask-tmp-packages-path))))))
+        (--first (s-match ".*\\.\\(tar\\|el\\)" it)
+                 (f-glob pattern cask-tmp-packages-path))))))
```
sambrightman: What changed here to require the additional matching? It looks a bit fragile at first glance but I lack context.

sambrightman:
I'll leave this another 24h for final comments and response to review comments because I'm tired and I've been away from Emacs development for a while. Ideally review comments are addressed but I think it's better to merge as-is than leave broken.

cask.el
tarsius: Since you don't want to call `package-build-archive` you have no other choice. I don't expect any future changes to the signature of this function but I haven't worked on this for a while so I am not sure. Maybe it will be made public again in the future - if it does, I hope I won't forget to add an alias.

cask.el
dajva: Some new (internal to `package-build`) file which has the same base (pattern) but different suffix is now added by `package-build`. This just makes sure that we chose the correct package file.

cask-bootstrap.el
dajva: Shouldn't it be required nevertheless?
My memory is a bit weak now but I think it was only emacs 24.3 that failed without this. Unfortunately I can't run this now locally for 24.3. I can do a fixup and see what travis says about it if you like?

dajva:
@sambrightman: Thanks for your comments.
I don't have much time to spend on this before the upcoming weekend. I can fix the minor issues then but I will not have time to do any refactoring related to wrapping `package-build` as tarsius suggest.
It would be helpful if you explicitly mark those issues that you want fixed before merging.
If you want to merge this as is asap, I can come back with another pr for the fixups. Just let me know.

cask-bootstrap.el
sambrightman: I thought that was due to melpa/package-build#19? I don't think eieio is available from any package repository anyway, and isn't used directly by cask?

sambrightman:
I clicked the wrong button for the review - with time in short supply, I agree it's better to tidy up later.

Issue #426, opened 2018-05-30
package-build-package is now private


GuiltyDolphin:
A [recent change](https://github.com/melpa/package-build/commit/7733c252691c2c1b8491c375176ca51a6e0dfd33#diff-ac4dc19e4f6618a58fcf1c0980e31857R751) to `package-build` renamed `package-build-package` to `package-build--package` (presumably to make it 'private').

This fails when specifying a fetcher, e.g., with the `Cask` file:

```emacs-lisp
(depends-on
  "cask"
  :git "git://github.com/cask/cask.git")
```

Running `cask install` gives:

```
Dependency cask failed to install: Symbol’s function definition is void: package-build-package
Output:

```

I'm not entirely sure what is intended to be used in place of this, but presumably `package-build-package` shouldn't be used any more.

rcoedo:
I'm having the same problem. Any workaround or suggestion?

GuiltyDolphin:
@rcoedo As a temporary work around, simply replacing references to `package-build-package` with `package-build--package` works. Looking at some of the other issues/pull requests that reference this, I think there may be a more permanent solution in the future.

sambrightman:
Fixed by #429 for now.

PR #429, opened 2018-06-04
Use private package-build--package

dajva:
This fixes issue #426.
`package-build-package` was renamed to private `package-build--package`.
Using the private function was recommended in https://github.com/melpa/package-build/issues/20 as a quick fix and to work with them to create a reasonable public API that cask can use.

Thanks


dajva:
Wonder if the 24.3 travis failures are related to packages abandoning support for that version of emacs?

Wilfred:
@sambrightman cask is currently broken without this -- would you mind merging?

sambrightman:
Sorry, have been rather busy in recent months. Test failure was a spurious network error from the look of it.

Issue #428, opened 2018-06-04
cask install no longer prints anything


Fuco1:
I'm not sure if this is my configuration somewhere but on none of my machines `cask install` prints anything.  I have no idea if it hangs and/or what is the progress.

Is there anything I can do?

sambrightman:
Have you tried `cask --help`? There are verbose and debug options.

Fuco1:
Any reason why it is not verbose by default?  I have not yet seen any package manager that does not print anything.  I would expect an opt-out switch like `-q` to be quiet instead.  I think this is an odd UX choice.

sambrightman:
It is common for Unix tools to be silent on success. You'd have to dig through the history to find any other reasoning for the change.

Fuco1:
This still does not work for me.

I've tried all the following 

```
cask --verbose install
cask --debug install
cask --verbose --debug install
cask install --verbose --debug
cask install --debug
cask install --verbose
```

It prints no progress except

```
Importing package-keyring.gpg...
Importing package-keyring.gpg...done
```

Can we try to add something that would at least say "Installing package X of Y"?  Every other package manager out there does that and it seems like a very standard feature.  If the bundle has 50 dependencies it is extremly confusing to me what's going on.

You said 

> It is common for Unix tools to be silent on success

which is true, but this is also very *uncommon* for interactive commands.  Imagine `grep` was silent on success.

PR #437, opened 2018-08-30
Log the install progress

Fuco1:
Related to #428 

https://asciinema.org/a/zKc2P9DOFyf2JknDTpA98noMN

We can make it a bit less stupid, this is a POC implementation.

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -41,7 +41,7 @@
   "Path to Cask bootstrap directory.")
 
 (defconst cask-bootstrap-packages
-  '(s dash f commander git epl shut-up cl-lib package-build eieio)
+  '(s dash f commander git epl shut-up cl-lib package-build eieio ansi)
   "List of bootstrap packages required by this file.")
 
 (unless (require 'package nil :noerror)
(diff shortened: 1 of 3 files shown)


rejeep:
Great stuff, I actually think this output should be visible even when not using the verbose flag. Don't know why I didn't want any output at first, but that was a mistake.

Fuco1:
Adding some colors, it could look like this for example

screen0152

rejeep:
Bam! 👊

rejeep:
@Fuco1 Should we perhaps merge this, kind of missing it...

Fuco1:
@rejeep I went through the patch again and there shouldn't be any changes to the logic, so this is safe to merge.

There are some minor details to be tweaked with the formatting but that can be done in a separate patch in the future. (for example now it assumes there's at most 99 packages, then the columns would mis-align).

rejeep:
Let's see how this works and we can add improvements later on. Nice work! 👍

rejeep:
I also just realised that it would be nice to have the same output for the update command.

Fuco1:
I'm now installing a new computer with 142 packages and this is really really nice.  I have a patch lined up to do the same thing with update.  Actually, just cleaning up the logs to be even prettier.

So thanks for pulling the trigger!

PR #443, opened 2018-09-01
Automatically initialize elpa dir when linking

### LINKED ISSUE #442: cask link does not work if cask install was not called before that.
We should be able to link a dependency right away, for example if there are no other dependencies we can't link at all.

### LINKED ISSUE #444: cask install should install dependencies of linked packages
I have a project with

```
> cask link list                                                
elsa-0.1.0          /home/matus/.emacs.d/projects/Elsa/
```

but when I

```
cask emacs -nw -q
```

and `(require 'elsa)` in scratch I get `"package not found: dash"`

I think we should install dependencies of linked packages.  I would not try to add the linked package's elpa dir to the loadpath as that can get quite messy (plus you have no idea where the pacakges load from if multiple linked dependencies reference the same dependency).

Fuco1:
Fixes #442, Fixes #444 

@rejeep Any reason why we did not do this?

I want to `cask init`, then add a `(depends-on <local-project>)` then `cask link <local-project> path` and it should work.

Instead we guarded essentially that `cask install` was run beforehand.  I don't think it's necessary but I'm not sure.

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -515,18 +515,49 @@ returns an `epl-package' object."
                    package-function))))))
      dependencies))))
 
+(defun cask--dependency-to-epl-package (dependency)
+  "Convert cask DEPENDENCY to `epl-package'."
+  (epl-package-create
+   :name (cask-dependency-name dependency)
+   :description
+    (if (fboundp 'package-desc-name)
+        (package-desc-create
+         :name (cask-dependency-name dependency)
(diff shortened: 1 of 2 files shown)


Fuco1:
Ok, this change actually exposed another one which I captured in #444. I'm adding a fix for that to this PR as well

rejeep:
> @rejeep Any reason why we did not do this?

I don't remember, sorry. But this makes sense to me.

PR #449, opened 2018-10-10
Use strings instead of bytes for env vars and paths

### LINKED ISSUE #382: Cask could not be boostrapped. Windows install using command prompt
Attached is an image showing the output from command prompt. Any assistance with this problem would be much appreciated. 

I've also tried cygwin-64. 

image

tom-bowles:
Fixes #382 and #285. The use of environb is unnecessary and doesn't work on Windows. From the os module docs (https://docs.python.org/3/library/os.html):

> In Python, file names, command line arguments, and environment variables are represented using the string type. On some systems, decoding these strings to and from bytes is necessary before passing them to the operating system. Python uses the file system encoding to perform this conversion (see sys.getfilesystemencoding()).

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -43,9 +43,6 @@ CASK = os.path.realpath(os.path.abspath(sys.argv[0]))
 CASK_BIN_DIRECTORY = os.path.dirname(CASK)
 CASK_DIRECTORY = os.path.dirname(CASK_BIN_DIRECTORY)
 
-# Get the byte-string process environment
-ENVB = getattr(os, 'environb', os.environ)
-
 # Regular expression to extract the version number from emacs --version
 VERSION_RE = re.compile(r'^GNU Emacs (?P<version>\d+(?:\.\d+)*)$', re.MULTILINE)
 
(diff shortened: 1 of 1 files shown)


tom-bowles:
It looks like the use of environb here was added in this PR: https://github.com/cask/cask/pull/133. The solution applied there is not the correct one. In particular, this comment on the issue is not accurate:

> I deliberately did not decode, because we have no need here. After all, environment variables are also just byte strings on the system level, so all we really need to do, is bytes-in, bytes-out. Coding system conversion just opens up for trouble…

Environment variables (and paths, etc) are strings on Windows and byte strings on Linux. Python explicitly takes care to shield scripts from this OS-level difference by converting them to strings using sys.getfilesystemencoding(). The correct solution to the issue identified in that PR is to decode the output of the child process using sys.getfilesystemencoding() to obtain the path as a string,

I will update this PR to do this.

Fuco1:
@tom-bowles Thanks for looking into this.  My low-level knowledge of python is very limited, I use it mostly just for stats.

@jorgenschaefer It might be a stretch but do you remember what happened in #133? Is the solution outlined here relevant? I don't want to take much of your time so if you don't know off the top of your head that's fine.  In which case @tom-bowles I would still rather wait for someone else to review this code.

tom-bowles:
I'm also happier with a review :-)

And thanks for your work on Cask.

jorgenschaefer:
I'm afraid I have no recollection whatsoever :-D

conao3:
#499 is merged, I close this for now. Thanks for your contribution!

PR #459, opened 2019-03-05
Add epl-refresh before install.

zhangyiant:
epl-install-file cannot install depenencies if package-contents are ready.
Need to refresh it first.


conao3:
LGTM

PR #460, opened 2019-03-17
edit install documentation

pierwill:
- retitle section "OS X Installation"
- add note that adding .cask to PATH is not always necessary
- regularize section title capitalization

### DIFF
--- a/doc/guide/installation.rst
+++ b/doc/guide/installation.rst
@@ -15,7 +15,7 @@ not work with Emacs 23 and below, or with other flavours of Emacs, e.g. XEmacs.
    Windows support for Cask requires additional work (see
    :ref:`windows setup`).
 
-Manual installation
+Manual Installation
 ===================
 
 To install Cask, run the following command:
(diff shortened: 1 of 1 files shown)


doc/guide/installation.rst
```diff
@@ -47,10 +47,7 @@ Cask is available in Homebrew_, so OS X users can just use:
 
 .. _Homebrew: http://brew.sh/
 
-Setup
-=====
-
-Add Cask to your `$PATH`:
+You may need to add Cask to your `$PATH`:
```
Fuco1: Maybe we can add a note on how to check if you need to or not.  I'm not sure about OSX but on linux you can do `which cask` to check for the executable.

pierwill: How about something like:

"If the `cask` command is not available after running `brew install cask`, you may need to add Cask to your `$PATH`:"

pierwill: You know, actually, I'm not so sure about this. I'll have to look into how the brew installation works. I have a working installation and no `.cask/` in my home dir. Should this be

`export PATH="$HOME/[[[.emacsd/]]].cask/bin:$PATH"`?

Fuco1: I'm not a brew user so I can't help here.

PR #468, opened 2019-09-05
Don't use hard-coded (insecure) temp file paths.

juergenhoetzel:
Use secure temp file functions instead and also ensure the removal of temporary files
after package installation.

Fixes symlink attacks like:
```
ln -s  /home/juergen/tmp/ /tmp/cask/checkout
```

This commit also extends the existing cask-checkout-and-package-dependency function  to install the  built package so that the temp path of the temporary file path is not used across function boundaries and easy deletion is possible.

### DIFF
--- a/cask.el
+++ b/cask.el
@@ -190,15 +190,6 @@ the function `cask--with-environment'.")
   '(:git :bzr :hg :darcs :svn :cvs)
   "List of supported fetchers.")
 
-(defconst cask-tmp-path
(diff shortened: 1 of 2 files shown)


juergenhoetzel:
> LGTM. please rebase on current HEAD and resolve conflict.

unfortunately, the commit can not  be easily rebased, since the signature of the function has changed:

It would now return a path to a (non-existent) temporary file

PR #481, opened 2020-06-20
Add ~/.cask to git command

eviltofu:
Added `~/.cask` to the end of the git command so that the subsequent instructions for modifying the PATH environment also works for git cloning.


conao3:
Good catch! thanks.

PR #483, opened 2020-08-14
Avoid twice call to ‘package-initialize’ in emacs 27/28

gilbertwong96:
In emacs 27/28, the function `package-initialize` will be called by emacs automatically.

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -53,7 +53,7 @@
             package-alist
             package-archive-contents
             (package-user-dir cask-bootstrap-dir))
-        (package-initialize)
+        (unless package--initialized (package-initialize))
         (condition-case nil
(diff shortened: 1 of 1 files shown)


conao3:
Thanks. Why do you pass `t` as `package-initialize`'s 1st argument?

gilbertwong96:
Something happened on travis-ci. @conao3

conao3:
Yes, I know. But your change is reasonable.
One point, why do you pass t as package-initialize's 1st argument?

gilbertwong96:
> Yes, I know. But your change is reasonable.
> One point, why do you pass t as package-initialize's 1st argument?

The `t` is nonsense, thanks for your advice.

conao3:
I think this PR is reasonable, merge it.
@rejeep, Please revert if there are any issues.

PR #492, opened 2020-11-09
Vendor older versions package-build & package-recipe for Emacs 24.*

immerrr:
This PR adds vendoring for earlier versions of `package-build` package that is compatible with Emacs 24.

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -47,6 +47,13 @@
 (unless (require 'package nil :noerror)
   (require 'package (expand-file-name "package-legacy" cask-directory)))
 
+(when (version< emacs-version "25.1")
+  ;; Use vendored package-build package (and package-recipe) because its newer
+  ;; versions require Emacs25.1+
+  (require 'package-recipe (expand-file-name "package-recipe-legacy" cask-directory))
+  (require 'package-build (expand-file-name "package-build-legacy" cask-directory)))
+
+
 (let ((orig-load-path load-path))
   (unwind-protect
       (let (package-archives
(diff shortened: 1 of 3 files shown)


cask-bootstrap.el
```diff
@@ -47,25 +47,36 @@
...
+                 (require package))
+             (error
+              (unless package-archives
+                (add-to-list 'package-archives (cons "gnu" "https://elpa.gnu.org/packages/"))
+                (add-to-list 'package-archives (cons "melpa" "https://stable.melpa.org/packages/"))
+                (package-refresh-contents))
+              (message "installing package %S" package)
+              (package-install package)
+              (require package))))
+         cask-bootstrap-packages))
```
conao3, line 79: Please revert this hunk to clear your change. If you want this change, please create another PR.

immerrr: I am happy to revert the "messages" part, but the workflow where every package is checked and installed one-by-one is important. Previous behaviour was to install all the missing packages if one of the requires fails. Without this change cask will not find some other bootstrap package and still try to install `package-build` along with the others.

immerrr: Ok, let me open a PR with just the change in logic and see if it works better.

conao3: > I am happy to revert the "messages" part, but the workflow where every package is checked and installed one-by-one is important. Previous behaviour was to install all the missing packages if one of the requires fails. Without this change cask will not find some other bootstrap package and still try to install package-build along with the others.

Nice. It's reasonable.

conao3: There are separated PR, #493

conao3:
Could you rebase, please?

immerrr:
Done!

conao3:
Thanks! merge.

PR #493, opened 2020-11-09
Check and install bootstrap packages one by one

immerrr:
This opens a path to vendoring bootstrap packages: they will not be installed
properly, but they will be requirable with some path extensions.

Pre-requisite to #492

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -54,18 +54,18 @@
             package-archive-contents
             (package-user-dir cask-bootstrap-dir))
         (package-initialize)
-        (condition-case nil
-            (mapc 'require cask-bootstrap-packages)
-          (error
-           (add-to-list 'package-archives (cons "gnu" "https://elpa.gnu.org/packages/"))
-           (add-to-list 'package-archives (cons "melpa" "https://melpa.org/packages/"))
-           (package-refresh-contents)
-           (mapc
-            (lambda (package)
-              (unless (package-installed-p package)
-                (package-install package)))
-            cask-bootstrap-packages)
-           (mapc 'require cask-bootstrap-packages))))
+        (mapc
+         (lambda (package)
+           (condition-case nil
+               (require package)
+             (error
+              (unless package-archives
+                (add-to-list 'package-archives (cons "gnu" "https://elpa.gnu.org/packages/"))
+                (add-to-list 'package-archives (cons "melpa" "https://stable.melpa.org/packages/"))
+                (package-refresh-contents))
+              (package-install package)
+              (require package))))
+         cask-bootstrap-packages))
     (setq load-path orig-load-path)))
 
 (provide 'cask-bootstrap)


cask-bootstrap.el
```diff
@@ -54,18 +54,19 @@
...
-            (lambda (package)
-              (unless (package-installed-p package)
-                (package-install package)))
-            cask-bootstrap-packages)
-           (mapc 'require cask-bootstrap-packages))))
+        (mapc
+         (lambda (package)
+           (condition-case nil
+               (progn
+                 (require package))
```
conao3, line 61: Please remove this unneeded `progn`?

cask-bootstrap.el
```diff
@@ -54,18 +54,19 @@
...
+             (error
+              (unless package-archives
```
conao3, line 63: Why check `package-archives` before modifying it?

cask-bootstrap.el
immerrr: Done

cask-bootstrap.el
immerrr: So that `package-archives` modification happens only once, because now `error` handler will be called for every package failed to be `required`.

conao3: good, but `package-archives` has ELPA entry as init value?
```
package-archives
;;=> (("gnu" . "https://elpa.gnu.org/packages/"))
```

immerrr: It is let-bound to `nil` for this block:

https://github.com/cask/cask/blob/88a89d9d4bb6e17f6ab470c348bd5698a3c2a3b8/cask-bootstrap.el#L52

conao3: I overlooked, thanks. So, please let bound as you expected in let 1st argument.

immerrr: I am sorry, I am not following. Right now it also serves a second purpose of a `package-initialized-p` variable to ensure we only fetch package contents once. Do you want me to change it?

conao3: I can't understand what you say "Right now it also serves a second purpose of a package-initialized-p variable to ensure we only fetch package contents once.", please tell me more?
And now your code does `(package-refresh-contents)` multiply in every loops?

immerrr: No, this version of the code does not run refresh multiple times. During the first iteration that needs to install something `package-archives` is `nil`, that block performs the initialization and populates `package-archives` list, so the subsequent iterations never enter the initialization block.

That is what I mean when I say it serves the second purpose of keeping "was initialized" state.

immerrr: @conao3 do you need something else from me here?

conao3: Sorry late reply, I was very busy recently. Nothing needed, thanks for your contribution!

conao3: And thank you for your additional description! I finally understood your code is everything ok.

conao3:
Thanks for your cooperation! And congrats on your first contribution to Cask! :tada:

cask-bootstrap.el
immerrr: Np, thank you for looking after cask!

PR #499, opened 2021-02-01
Add shim to workaround environb missing on Windows.

cniles:
Possibly a duplicate of https://github.com/cask/cask/pull/449; I coded it up and tested it before checking open pulls (though that PR is few years old).

This PR shims environ to decode bytes strings or convert other params to strings for Windows.  Tested on Windows/Linux with python 2 and 3.

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -36,6 +36,7 @@ import os
 import re
 import subprocess
 import errno
+import platform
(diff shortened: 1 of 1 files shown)


conao3:
Thanks, good catch!

PR #500, opened 2021-02-02
Silence ‘package-initialize’ warning on startup

alexjgriffith:
### Issue
The `package-initialize` call in `cask-bootstrap.el` is necessary for Cask to function properly, and since the introduction of macs 27, a warning has been thrown whenever `package-initialize` is called.

```
Warning (package): Unnecessary call to ‘package-initialize’ in init file [2 times]
```

### Proposed Solution
For versions 27+, set `package--initialized` to nil immediately before `package-initialize` is called in `cask-bootstrap.el`, to silence this warning. 

### Effects
This only short circuits following in `package-initialize`. `package--initialized` is set to `'t` before the end of `package-initialize` and there are no other conditional checks that reference it before that point (including in called functions).

``` elisp
  (when (and package--initialized (not after-init-time))
    (lwarn '(package reinitialization) :warning
           "Unnecessary call to `package-initialize' in init file"))

```

### DIFF
--- a/cask-bootstrap.el
+++ b/cask-bootstrap.el
@@ -60,6 +60,8 @@
             package-alist
             package-archive-contents
             (package-user-dir cask-bootstrap-dir))
+        (when (not (version< emacs-version "27"))
+            (setq package--initialized nil))
         (package-initialize)
         (mapc
          (lambda (package)


cask-bootstrap.el
```diff
@@ -60,6 +60,8 @@
             package-alist
             package-archive-contents
             (package-user-dir cask-bootstrap-dir))
+        (when (not (version< emacs-version "27"))
```
jcs090218, line 63: Can you replace `when` and `not` to just `unless`?

jcs090218 (approved):
LGTM.

dickmao:
I'm under the assumption https://github.com/cask/cask/blob/6fe0a7230ced9c6e91c94a91bc6359f5e4b60545/cask-bootstrap.el#L48 is the first and only invocation of `package-initialize`.  I trust you will let me know if I'm mistaken.  Thanks.

PR #511, opened 2021-02-09
Work again on Emacs-24.5

conao3:
Work again on Emacs-24.5


conao3:
It because setup-cask uses cask HEAD so my patch is not included in the Cask...
Bootstrap problem is hard...

PR #512, opened 2021-02-10
More support for emacs 24.4

immerrr:
This should fix couple of remaining issues that were in the way of supporting Emacs 24.4:

- it forces the use of external `openssl` client to workaround the gnutls bug that was fixed in Emacs 24.5
- it adds `cl-generic` backport to provide the missing cl macros


conao3:
Thanks! I tried it and it works correctly!

PR #525, opened 2021-03-05
Fix environment variable values as well.

cniles:
Fixes issue with environment variables in Windows not supporting bytes
objects for variable keys or values.  An error will occur when running
'cask emacs', this corrects that problem.


conao3:
This seems good (I don't have a Windows, I can't test it), thanks!

Issue #523, opened 2021-02-19
Drop Emacs-24.1~24.4 support?


conao3:
Emacs-24.4 is not able to download packages from MELPA. https://github.com/Silex/docker-emacs/issues/72
[emacs-nix-ci](https://github.com/purcell/nix-emacs-ci) can be downloaded by purcel's patch, but I'm bit hesitated about testing it with patched Emacs.
(Even if CI can be forced to work, if it doesn't work in the user's environment, there is no reason to support it.)

However, cask is a widely used build tool, and raising the required version of cask requires a cautious attitude.
But even now, Emacs-24.1, 24.2, and 24.3 cannot resolve the cask dependency and CI is failing. https://github.com/cask/cask/actions/runs/566090457 (CI on Emacs-25 is failed randomly for other reason)

[Tarsius's point](https://github.com/cask/cask/issues/491#issuecomment-724840312) is also in a joking tone, but it seems to good point I think.

CC:
- cask author: @rejeep 
- MELPA maintainers: @milkypostman, @purcell, @tarsius, @riscy

---
Resources
- List of version (https://repology.org/project/emacs/versions)
  - Emacs-24.1
  - Emacs-24.2
  - Emacs-24.3: CentOS 7 (EOL at 2024), Ubuntu 14.04 (EOL at 2022)
  - Emacs-24.4
  - Emacs-24.5: Ubuntu 16.04 (EOL at 2024)
  - Emacs-25.1
  - Emacs-25.2: Ubuntu 18.04 (EOL at 2028)
  - Emacs-25.3

- EOL info
  - https://wiki.ubuntu.com/Releases
  - https://wiki.centos.org/About/Product

dickmao:
I believe I've resolved the issue in commit 37052df, in which I determined 24.1 and 24.2, because they lack the function definition of `cl-flet`, would be a bridge too far for continued testing on those versions.  Currently `master` tests from 24.3 onwards.

I won't close this issue after you've inexplicably reopened it, except to say inexperienced devs erroneously consider issue proliferation as progress, when in fact adding labels and issues do nothing to advance the project (because no code was written).

conao3:
Writing code isn't the only way forward.
As far as I can tell from the EOL information and the new release information in the distribution, I think it's fair to start supporting Emacs-24.5. However, we don't yet know how many packages this decision will affect. We don't have enough information.

This project was led by @rejeep and we just assumed his project. We need to be humble in our maintenance.

dickmao:
[...]

PR #536, opened 2021-03-11
Green check marks

dickmao:
We should strive for green check marks.


.github/workflows/test.yml
```diff
@@ -1,52 +1,38 @@
...
+    - '**.rst'
+    branches-ignore:
+    - 'master'
+    - 'main'
 
 jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
```
conao3, line 10: Please keep this fail-fast config? Please tell me a reason to remove this?

.github/workflows/test.yml
```diff
@@ -1,52 +1,38 @@
...
-          - '26.2'
-          - '26.3'
-          - '27.1'
-          - 'snapshot'
-        include:
-          - emacs_version: 'snapshot'
-            allow_failure: true
+        os: [ubuntu-latest]
+        python_version: [3.6]
+        emacs_version: [24.4, 24.5, 25.3, 26.3, 27.1]
```
conao3, line 23: Please check every Emacs-24.1~Emacs27.1 and snapshot.
If you want to remove this (I think so too), we should finish the #523 discussion.

.github/workflows/test.yml
dickmao: Restored 37052df

.github/workflows/test.yml
dickmao: Addressed 37052df

conao3:
It doesn't make much sense to keep it like this, so I'll merge it.
Well, we haven't been able to test Emacs-24.1 and 24.2, but I feel it's fair to support only Emacs-24.5 and above.
Anyway, it seems that cask can be tested more robustly with your contribution. Thanks!

PR #543, opened 2021-03-30
Fix unbounded local variable

calancha:
### Descriptionr
I am observing in someTravis logs the following error:

cask-unbound-variable-Screenshot from 2021-03-30 12-51-28

It looks like an unbounded local variable.

### Suggested fix

* cask.el (cask-install): Signal the error as in `cask-update'.


dickmao:
My bad, thanks.

PR #545, opened 2021-04-02
Show deprecation notice if user run cask with Emacs <= 24.4

conao3:
Show deprecation notice if user run cask with Emacs <= 24.4

https://github.com/cask/cask/issues/523#issuecomment-809378637

### DIFF
--- a/cask-cli.el
+++ b/cask-cli.el
@@ -43,6 +43,9 @@
 (defconst cask-cli--table-padding 10
   "Number of spaces to pad with when printing table.")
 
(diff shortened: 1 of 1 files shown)


conao3:
Screenshot_2021-04-03_00-28-09

I add `--no-emacs-24-warning` option. If users want to hide this warning, the user can use this option.

(To be honest, I can't run `cask` (`package.el`) with Emacs <= 24.4, I take this screenshot with Emacs-24.5 (with temporary modified source))

PR #547, opened 2021-04-04
Ensure CI is run when pushing to master

### LINKED ISSUE #539: GHA badge is not updated?
This is not important thing, but our CI is [green](https://github.com/cask/cask/actions) but badge on README is red which I see.
Do anyone know why?

akirak:
Fixes #539.

I excluded `paths-ignore`, because I think that filter has little impact, because you don't update the documentation as frequently as the code. If you still want them, you could change it as follows, because README is written in Org:

```yaml
    paths-ignore:
    - '**.md'
    - '**.org'
    - '**.rst'
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,16 +2,9 @@ name: CI
 
 on:
   pull_request:
-    paths-ignore:
-    - '**.md'
-    - '**.rst'
   push:
-    paths-ignore:
-    - '**.md'
-    - '**.rst'
-    branches-ignore:
+    branches:
     - 'master'
-    - 'main'
 
 jobs:
   build:


jcs090218:
LGTM

dickmao:
The changes in #548 are excellent.

The changes here in #547 I am less enthused about.

The idea generally should be to minimize the number of times we "roll the dice" with GHA because tests are wont to fail for transient reasons upstream (cachix, azure, keyservers).  In industry parlance, this is known as "flapping."

If a single gha line item has a 1% chance of flapping, and we currently have seven line items, then a PR has 0.99^7 = 93% of success.   With this change, we roll the dice 7 more times, so success rate drops to 87%.

Three people currently can push to master.  I would think all three would PR to our own fork, get green check marks, then push directly to master.  This is better both for our sanity and climate change :wink:.

akirak:
@dickmao 

> If a single gha line item has a 1% chance of flapping, and we currently have seven line items, then a PR has 0.99^7 = 93% of success. With this change, we roll the dice 7 more times, so success rate drops to 87%.

I don't know what this means, but please feel free to restore the `paths-ignore` settings if you find them important.

dickmao:
It's not the `paths-ignore` I'm worried about (we should restore those as well since it's pointless to rerun gha on doc changes).  I find greater fault in removing the branch-ignore on push master.

Have you not ever PR'ed to your own repo, passed all the tests, then merged the PR and groaned because github pointlessly ran all the tests again?

akirak:
@dickmao 

> github pointlessly ran all the tests again?

Did you see the issue #539? It is necessary for updating the status badge.

If you want to skip all the tests on master, you could remove the entire `push` setting. However, the badge makes no sense if the status is not updated at all. Why did you add the badge?

> we should restore those as well since it's pointless to rerun gha on doc changes

I know that. It depends on how frequently you update the docs and how much you care about the Actions usage. If you don't update docs often, it won't make much difference, so I have removed the settings for conciseness. But I don't mind at all if you restore the settings.

dickmao:
The badge, I admit is rather pointless, and certainly not worth maintaining if it means we have to run twice as many tests as necessary.  If we restore the branch-ignore on master, the badge should remain green in perpetuity, which ironically, fully realizes the purpose of the badge, which is ostentation not utility.

Economy of expression (conciseness) is something I value very much.  But I equate the doc ignore's to say adding `# -*- coding: utf-8 -*-` to a python file.  Its functional value is *de minimus*, but it's just good form.

In any case, I want to thank you for your contributions, in particular #548, and will leave as-is.  Just don't want anybody to feel slighted should I choose to undo #547 later.

akirak:
@dickmao Okay, I am not a maintainer of this project, and I am even not using cask, so please go ahead. Thanks for your explanation.

PR #548, opened 2021-04-04
Clean up the CI configuration using matrix.include

akirak:
I think this is cleaner than putting `if` clause in every step. See [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations) on the syntax.


jcs090218 (approved):

jcs090218:
Thanks! :)

PR #549, opened 2021-04-04
CI badge display on master

jcs090218:
Correct path for CI badge.

### DIFF
--- a/README.org
+++ b/README.org
@@ -2,7 +2,7 @@
 [[file:cask.png]]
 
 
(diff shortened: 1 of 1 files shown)


conao3:
Just question, why we need this? I haven't used this feature (URL query), my GHA badge works correctly.

jcs090218:
I don't have the correct answer but by having this argument will cause the CI badge stays in `no status`. This PR only revert and display the master branch's svg which I think it's the standard from GHA.

conao3:
Oh, your PR remove the URL query?
I have never used this query before, so I am not sure that option is necessary.

conao3 (approved):

PR #558, opened 2021-12-19
Fix shebang

berquist:
The fix from #14 got lost when `bin/cask` was first translated to Python and then translated back in d1ee01ba. This puts it back so that Cask works with Nix and other platforms that don't have `bash` even symlinked in `/bin`.

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -1,4 +1,4 @@
-#!/bin/bash -eu
+#!/usr/bin/env -S bash -eu
 
 # Copyright (C) 2012, 2013, 2014 Johan Andersson
 # Copyright (C) 2013-2014, 2016 Sebastian Wiesner


dickmao:
Word.

akirak:
Note that `-S` option of env command is only available since coreutils 8.30. Ubuntu 18.04 doesn't support it, so some users may not be able to run the script.

berquist:
> Note that `-S` option of env command is only available since coreutils 8.30. Ubuntu 18.04 doesn't support it, so some users may not be able to run the script.

Well that's no good.  Is it enough to do ye olde `#!/usr/bin/env bash; set -euo pipefail; <script contents ...>`?

dickmao:
> Ubuntu 18.04 doesn't support it, so some users may not be able to run the script.

True that.  Reverted, thanks.

akirak:
> Well that's no good. Is it enough to do ye olde #!/usr/bin/env bash; set -euo pipefail; <script contents ...>

Yes, that will be suitable here.

dickmao:
Awesome, guys.  #14 restored in 751435b.  Sorry about that.

PR #561, opened 2022-03-17
Detect functionality, not OS

wyuenho:
`readlink -f` works on macOS 12, this PR makes Python unnecessary for those systems.

### DIFF
--- a/bin/cask
+++ b/bin/cask
@@ -38,12 +38,8 @@ case $subcommand in
 		;;
 	*)
 		WHICH=${WHICH:-which}
-		READLINK=$($WHICH readlink || true)
-		if [ "$(uname || true)" == "Darwin" ] ; then
-			READLINK=
-		fi
-		READLINK=${READLINK:-$($WHICH greadlink || true)}
-		if [ ! -z "$READLINK" ] ; then
+		READLINK=$($WHICH readlink || $WHICH greadlink)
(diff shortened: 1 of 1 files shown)


bin/cask
```diff
@@ -38,12 +38,8 @@ case $subcommand in
...
 	*)
 		WHICH=${WHICH:-which}
-		READLINK=$($WHICH readlink || true)
-		if [ "$(uname || true)" == "Darwin" ] ; then
-			READLINK=
-		fi
-		READLINK=${READLINK:-$($WHICH greadlink || true)}
-		if [ ! -z "$READLINK" ] ; then
+		READLINK=$($WHICH readlink || $WHICH greadlink)
+		if [ -n "$READLINK" ] && eval "$READLINK" -f "$CASK" &>/dev/null; then
```
dickmao, line 42: On an old apple, it'll presumably try `readlink -f`, fail, then python it up.  The old code would have tried `greadlink`.  Ideally, we would try `readlink -f`, fail, try `greadlink`, fail, then python it up.

Maybe `READLINK=$($WHICH greadlink || $WHICH readlink)` would be cleanest.

wyuenho: What do you think of using [this](https://stackoverflow.com/a/18443300/226114) to remove python completely?

PR #595, opened 2023-10-02
Use python3

wyuenho:
python doesn't usually come preinstalled on any recent OSes anymore, but python3 does.


jcs090218 (approved):

wyuenho:
Sorry, no need for this PR. This was an attempt to use the correct python in older macOS, but turns out macOS didn't come with python3 until 12.4, and since then readlink -f works.

PR #605, opened 2025-08-01
Add gnu-devel source

wyuenho:
Packages that depend on packages now only developed on the Emacs main tree such as eglot and use-package may occasionally benefit from this.


jcs090218 (approved):

jcs090218:
Thank you!

